Issue #3, opened 2022-05-10
Remove operator specific code


Jaanki:
- [x] In particular, bundle, config, controllers, deploy, packagemanifests, package/*-operator*, scripts/operator*, bundle.Dockerfile directories/files.
- [x]  Change module name in go.mod to github.com/submariner-io/subctl and add a dependency on submariner-operator.
- [x]  Delete e2e-all-k8s.yml CI jobs.

Part of [Epic: Move subctl to its own repo](https://github.com/submariner-io/enhancements/issues/89#)

PR #4, opened 2022-05-10
Remove building/releasing operator image and bundle

tpantelis:
Also remove the operator controller code.

Related to https://github.com/submariner-io/subctl/issues/3.


skitt (changes requested):
The license check needs to be updated to check `subctl` instead of `submariner-operator`.

skitt (approved):

PR #5, opened 2022-05-10
Use embeddedyamls and Submariner CRD APIs from submariner-operator

tpantelis:
Related to https://github.com/submariner-io/subctl/issues/3.


skitt (approved):

cmd/subctl/deploybroker.go
```diff
@@ -24,13 +24,13 @@ import (
...
-	"github.com/submariner-io/submariner-operator/internal/exit"
-	"github.com/submariner-io/submariner-operator/pkg/broker"
-	"github.com/submariner-io/submariner-operator/pkg/client"
-	"github.com/submariner-io/submariner-operator/pkg/deploy"
+	"github.com/submariner-io/subctl/internal/cli"
+	"github.com/submariner-io/subctl/internal/component"
+	"github.com/submariner-io/subctl/internal/constants"
+	"github.com/submariner-io/subctl/internal/exit"
+	"github.com/submariner-io/subctl/pkg/broker"
+	"github.com/submariner-io/subctl/pkg/client"
```
Jaanki, line 32: Lets reuse this from operator itself.

Jaanki: Seems commit [b34d651](https://github.com/submariner-io/subctl/pull/5/commits/b34d651dc745478bf1650ad45f1a4a8019b2c2d8) was pushed while I was reviewing.

tpantelis: That has the client `Producer` which is only used by `subctl`. The controllers use the `sigs.k8s.io/controller-runtime/pkg/client` interface.

Jaanki: `pkg/broker` , `pkg/diagnose` (and may be few others) do need it and we might end up keeping these packages in `operator` repo (because controllers need them). Lets keep it here at the moment and see how things turn out.

Jaanki (approved):

Jaanki (approved):

cmd/subctl/deploybroker.go
tpantelis: `pkg/diagnose` is not used by the controllers but parts of `pkg/broker` is so you're right, we need to keep the client `Producer` in submariner-operator. I'll change it.

PR #13, opened 2022-05-12
Use new subctl implementation

### LINKED ISSUE #10: Use new subctl implementation
Use subctl from `cmd` and not `pkg/subctl`

Jaanki:
1. Don't build old subctl in Makefile
2. Remove pkg/subctl/cmd,benchmark packages
3. move pkg/subctl/table -> internal/show/table
4. move pkg/subctl/cmd/utils -> internal/utils

Closes: https://github.com/submariner-io/subctl/issues/10


internal/utils/utils.go
```diff
@@ -20,52 +20,17 @@ package utils
...
-}
-
-// ExpectFlag exits with an error if the flag value is empty.
-func ExpectFlag(flag, value string) {
-	if value == "" {
-		ExitWithErrorMsg(fmt.Sprintf("You must specify the %v flag", flag))
-	}
-}
-
 func GetSubmarinerResourceWithError(config *rest.Config) (*v1alpha1.Submariner, error) {
```
Jaanki: I think there was another function to get `submariner`. Can't find it. @tpantelis Do you remember it?

Makefile
```diff
@@ -124,37 +108,37 @@ cmd/bin/subctl-%: $(shell find cmd/ -name "*.go") $(VENDOR_MODULES)
 	GOARCH=$${components[-1]}; \
 	export GOARCH GOOS; \
 	$(SCRIPTS_DIR)/compile.sh \
-		--ldflags "-X 'github.com/submariner-io/submariner-operator/pkg/version.Version=$(VERSION)' \
+		--ldflags "-X 'github.com/submariner-io/subctl/pkg/version.Version=$(VERSION)' \
 		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}'" \
         --noupx $@ ./cmd $(BUILD_ARGS)
 
-ci: golangci-lint markdownlint unit build images
```
skitt, line 131: Why are these removed?

.github/workflows/e2e-new-subctl.yml
```diff
@@ -13,10 +13,6 @@ jobs:
       - name: Check out the repository
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
-      - name: make old subctl
```
skitt, line 16: This whole file should be removed.

skitt (changes requested):

.github/workflows/e2e-new-subctl.yml
tpantelis: Perhaps we should keep this and remove the other tests from the [Makefile](https://github.com/submariner-io/subctl/blob/3e7bae502dec0347f8516b823f0eb00d9a29da9c/Makefile#L134). I like the output organized in separate expandable sections. Unless there's a specific reason we want them in the _Makefile_?

Makefile
```diff
@@ -124,37 +108,37 @@ cmd/bin/subctl-%: $(shell find cmd/ -name "*.go") $(VENDOR_MODULES)
...
+ci: markdownlint build images
 
 # Test as many of the config/context-dependent subctl commands as possible
-test-subctl: bin/subctl deploy
+test-subctl: cmd/bin/subctl deploy
```
tpantelis, line 118: We should either go with this or the GHA. I prefer the latter.

Makefile
Jaanki: Becasue these targets were removed in the earlier commit https://github.com/submariner-io/subctl/pull/5/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L177.

.github/workflows/e2e-new-subctl.yml
Jaanki: We were discussing this yesterday and hence I haven't removed this file.

internal/utils/utils.go
tpantelis: AFAICT, this function is now only used by `restconfig.CheckVersionMismatch` so we can simply inline the code there - it's  not much. Plus, as you probably know, I'm not a fan of `utils` packages 😄 

```Go
func (rcp *Producer) CheckVersionMismatch(cmd *cobra.Command, args []string) error {
	...
	operatorClient, err := subOperatorClientset.NewForConfig(config.Config)
	exit.OnErrorWithMessage(err, "Error creating operator clientset")

	submariner, err := operatorClient.SubmarinerV1alpha1().Submariners(constants.OperatorNamespace).Get(
		context.TODO(), constants.SubmarinerName, v1.GetOptions{})
	if apierrors.IsNotFound(err) {
		return nil
	}

	exit.OnErrorWithMessage(err, "Error obtaining Submariner resource")

	if submariner.Spec.Version != "" {
		...
	}

	return nil
}
```

tpantelis (changes requested):
_pkg/subctl/datafile_ should be removed.

_pkg/subctl/operator_ is still used but I can refactor that in another PR.

Makefile
tpantelis: These targets still exist in shipyard - we just removed the additional dependent targets in this file. So they should remain on this line although I'm not sure what this target is used for.

Jaanki:
> _pkg/subctl/datafile_ should be removed.
> 
> _pkg/subctl/operator_ is still used but I can refactor that in another PR.

My focus is only on `pkg/subctl/cmd` and packages it affects. The other 2 weren't looked upon. Once we start using new `subctl` and it is built, work on operator upgrading can be started.

.github/workflows/e2e-new-subctl.yml
skitt: Having it in the Makefile makes it easy to run locally as well as in CI. If you want expandable sections, the Makefile can be structured to provide those.

tpantelis (approved):

skitt (approved):

PR #66, opened 2022-05-25
Build subctl images for x86-64 and ARM64

skitt:

### DIFF
--- a/.github/workflows/multiarch.yml
+++ b/.github/workflows/multiarch.yml
@@ -1,0 +1,19 @@
+---
+name: Multi-arch Builds
+
+on:
+  pull_request:
+
+jobs:
+  check-multiarch:
+    name: Check the multi-arch builds
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - name: Set up QEMU (to support building on non-native architectures)
+        uses: docker/setup-qemu-action@8b122486cedac8393e77aa9734c3528886e4a1a8
+      - name: Build the multi-arch images
+        run: make multiarch-images
+      - name: Check that we actually build multi-arch images
+        run: bash -c '[ "$(echo package/*.tar)" != "package/*.tar" ]'
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,10 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 # Running in Dapper
 
+PLATFORMS ?= linux/amd64,linux/arm64
 IMAGES = subctl
 PRELOAD_IMAGES := $(IMAGES) submariner-operator submariner-gateway submariner-globalnet submariner-route-agent lighthouse-agent lighthouse-coredns nettest
+MULTIARCH_IMAGES := subctl
 undefine SKIP
 undefine FOCUS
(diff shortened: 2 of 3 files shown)


tpantelis (approved):

.github/workflows/multiarch.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+      - name: Set up QEMU (to support building on non-native architectures)
+        uses: docker/setup-qemu-action@8b122486cedac8393e77aa9734c3528886e4a1a8
+      - name: Build the multi-arch images
+        run: make multiarch-images
+      - name: Check that we actually build multi-arch images
+        run: bash -c '[ "$(echo package/*.tar)" != "package/*.tar" ]'
```
mkolesnik, line 19: BTW why is it being run in `bash -c`?
Is it a GHA quirk?

package/Dockerfile.subctl
```diff
@@ -1,8 +1,21 @@
...
+ARG BASE_BRANCH
+ARG SOURCE=/go/src/github.com/submariner-io/submariner-operator
+
+FROM --platform=${BUILDPLATFORM} quay.io/submariner/shipyard-dapper-base:${BASE_BRANCH} AS builder
+ARG SOURCE
+ARG TARGETPLATFORM
+
+COPY . ${SOURCE}
+
+RUN rm -f cmd/bin/subctl && make -C ${SOURCE} LOCAL_BUILD=1 cmd/bin/${TARGETPLATFORM}/subctl
```
mkolesnik:
```suggestion
RUN rm -f cmd/bin/subctl && make -C ${SOURCE} cmd/bin/${TARGETPLATFORM}/subctl
```

There's no significance to `LOCAL_BUILD` here since it's already inside the shipyard container env.

Makefile
```diff
@@ -100,6 +101,13 @@ dist/subctl-%.tar.xz: cmd/bin/subctl-%
 	tar -cJf $@ --transform "s/^cmd.bin/subctl-$(VERSION)/" $<
 
 # Versions may include hyphens so it's easier to use $(VERSION) than to extract them from the target
+
+# Special case for Linux container builds
```
mkolesnik, line 105: Is there some more info that can be provided here (eg link) for why this special case is needed?

It's not clear why you need to pre create the dir, nor why youre re-linking `cmd/bin/subctl` at the end (as it's not the target of this build)

.github/workflows/multiarch.yml
skitt: Yes, I couldn’t get the test to work using only GHA constructs. I could specify `sh -c` instead if you prefer.

Makefile
skitt: See the comment in the updated PR. I’m not sure what you mean by “re-linking `cmd/bin/subctl` at the end”; this is linking the requested target to the listed prerequisite.

dfarrell07 (approved):

Makefile
mkolesnik: OK gotcha, I thought it's overriding the link for `cmd/bin/subctl`

mkolesnik (approved):
Hopefully it won;t break the release ;)

PR #16, opened 2022-05-12
Add support for Azure

Jaanki:
Depends on https://github.com/submariner-io/cloud-prepare/pull/203

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

PR #74, opened 2022-05-31
Open metrics port for globalnet by default

### LINKED ISSUE #73: On OpenShift, none of the Globalnet metrics show up
I set up three clusters (one broker, two joined clusters) running OpenShift, and installed Submariner with Globalnet. This can be confirmed by looking at the Cluster objects:

```yaml
apiVersion: submariner.io/v1
kind: Cluster
metadata:
[...]
  name: skitt-1
  namespace: submariner-k8s-broker
  resourceVersion: '57012'
  uid: a84b25f2-35b7-4a84-a981-6ba889d9373a
spec:
  cluster_cidr:
    - 10.128.0.0/14
  cluster_id: skitt-1
  color_codes:
    - blue
  global_cidr:
    - 242.0.0.0/16
  service_cidr:
    - 172.30.0.0/16
```

However, none of the Globalnet metrics show up in OpenShift monitoring:

Screenshot from 2022-03-04 14-22-06

maayanf24:
1) allow metrics-port flag for cloud prepare cmd to accept
multiple values
2) make globalnet metrics port to be default as well

Fixes: #73

Signed-off-by: Maayan Friedman <maafried@redhat.com>

### DIFF
--- a/cmd/subctl/cloud.go
+++ b/cmd/subctl/cloud.go
@@ -58,12 +58,14 @@ func init() {
 	restConfigProducer.AddKubeContextFlag(cloudCmd)
 	rootCmd.AddCommand(cloudCmd)
 
+	cloudPorts.Metrics = append(cloudPorts.Metrics, 8080, 8081)
+
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Natt, "natt-port", port.ExternalTunnel,
 		"IPSec NAT traversal port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.NatDiscovery, "nat-discovery-port",
 		port.NATTDiscovery, "NAT discovery port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Vxlan, "vxlan-port", port.IntraClusterVxLAN, "Internal VXLAN port")
-	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Metrics, "metrics-port", 8080, "Metrics port")
+	cloudPrepareCmd.PersistentFlags().Var(&cloudPorts.Metrics, "metrics-port", "Metrics port")
 
 	cloudCmd.AddCommand(cloudPrepareCmd)
 
--- a/cmd/subctl/cloud.go
+++ b/cmd/subctl/cloud.go
@@ -19,6 +19,9 @@ limitations under the License.
 package subctl
 
 import (
+	"strconv"
+
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/submariner-io/subctl/pkg/cloud"
 	"github.com/submariner-io/submariner/pkg/port"
@@ -58,16 +61,41 @@ func init() {
 	restConfigProducer.AddKubeContextFlag(cloudCmd)
 	rootCmd.AddCommand(cloudCmd)
 
-	cloudPorts.Metrics = append(cloudPorts.Metrics, 8080, 8081)
-
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Natt, "natt-port", port.ExternalTunnel,
 		"IPSec NAT traversal port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.NatDiscovery, "nat-discovery-port",
 		port.NATTDiscovery, "NAT discovery port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Vxlan, "vxlan-port", port.IntraClusterVxLAN, "Internal VXLAN port")
-	cloudPrepareCmd.PersistentFlags().Var(&cloudPorts.Metrics, "metrics-port", "Metrics port")
+
+	cloudPorts.Metrics = append(cloudPorts.Metrics, 8080, 8081)
+
+	cloudPrepareCmd.PersistentFlags().Var(&uint16Slice{value: &cloudPorts.Metrics}, "metrics-ports", "Metrics ports")
+
+	cloudPrepareCmd.PersistentFlags().Var(&metricsAliasType{}, "metrics-port", "Metrics port")
+	_ = cloudPrepareCmd.PersistentFlags().MarkDeprecated("metrics-port", "Use metrics-ports instead")
 
 	cloudCmd.AddCommand(cloudPrepareCmd)
 
 	cloudCmd.AddCommand(cloudCleanupCmd)
+}
+
+type metricsAliasType struct{}
+
+func (m metricsAliasType) String() string {
+	return strconv.FormatUint(uint64(cloudPorts.Metrics[0]), 10)
+}
+
+func (m metricsAliasType) Set(s string) error {
+	v, err := strconv.ParseUint(s, 0, 16)
+	if err != nil {
+		return errors.Wrap(err, "conversion to uint16 failed")
+	}
+
+	cloudPorts.Metrics = []uint16{uint16(v)}
+
+	return nil
+}
+
+func (m metricsAliasType) Type() string {
+	return "uint16"
 }
--- a/cmd/subctl/uint16slice.go
+++ b/cmd/subctl/uint16slice.go
@@ -16,11 +16,41 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package cloud
+package subctl
 
-type Ports struct {
-	Natt         uint16
-	NatDiscovery uint16
-	Vxlan        uint16
-	Metrics      []uint16
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type uint16Slice struct {
+	value *[]uint16
+}
+
+func (s *uint16Slice) Type() string {
+	return "uint16Slice"
+}
+
+func (s *uint16Slice) String() string {
+	return fmt.Sprintf("%v", *s.value)
+}
+
+func (s *uint16Slice) Set(value string) error {
+	values := strings.Split(value, ",")
+
+	*s.value = make([]uint16, len(values))
+
+	for i, d := range values {
+		u, err := strconv.ParseUint(d, 10, 16)
+		if err != nil {
+			return errors.Wrap(err, "conversion to uint16 failed")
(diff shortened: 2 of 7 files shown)


pkg/cloud/cloud.go
```diff
@@ -18,9 +18,54 @@ limitations under the License.
...
 
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type uintslice []uint16
```
Jaanki: unit16slice?

pkg/cloud/cloud.go
```diff
@@ -18,9 +18,54 @@ limitations under the License.
...
+func (i *uintslice) Contains(value uint16) bool {
+	for _, port := range *i {
+		if value == port {
+			return true
+		}
+	}
+
+	return false
+}
+
```
Jaanki: Where are these used?

maayanf24: in struct `Ports.Metrics` to allow multiple ports to be pass by  flag `metrics-ports`

cmd/subctl/cloud.go
```diff
@@ -57,10 +57,12 @@ func init() {
 	restConfigProducer.AddKubeContextFlag(cloudCmd)
 	rootCmd.AddCommand(cloudCmd)
 
+	cloudPorts.Metrics = append(cloudPorts.Metrics, 8080, 8081)
+
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Natt, "natt-port", 4500, "IPSec NAT traversal port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.NatDiscovery, "nat-discovery-port", 4490, "NAT discovery port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Vxlan, "vxlan-port", 4800, "Internal VXLAN port")
-	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Metrics, "metrics-port", 8080, "Metrics port")
+	cloudPrepareCmd.PersistentFlags().Var(&cloudPorts.Metrics, "metrics-port", "Metrics port")
```
tpantelis: Can't we use `UintSliceVar` instead of defining our own flag type?

```suggestion
	cloudPrepareCmd.PersistentFlags().UintSliceVar(&cloudPorts.Metrics, "metrics-port", "Metrics port")
```

tpantelis (changes requested):

Jaanki:
@maayanf24 since, https://github.com/submariner-io/subctl/pull/16 is merged, the same changes would also need to be done for Azure.

cmd/subctl/cloud.go
Jaanki: IIRC, that needs [uint](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.UintSliceVar) which should be [atleast 32 bits in size.](https://pkg.go.dev/builtin#uint). Ours is unit16.

pkg/cloud/cloud.go
Jaanki: Got it. These methods are needed by the `Value` Type in `Var`

maayanf24:
> @maayanf24 since, #16 is merged, the same changes would also need to be done for Azure.

sure, I'll add them

pkg/cloud/cloud.go
skitt: This bothers me: `cloud.Ports` is part of our Cobra-independent API, but we’re allowing Cobra-related concerns to leak into it here. We should change this to a `[]uint16` if that’s what we need, or a `[]uint` to make things simpler (since we’re changing all the metrics-port-handling code in `pkg/cloud/prepare` anyway, the latter isn’t a problem in practice, although it does mean more leaking of Cobra concerns into `pkg/cloud` albeit at a smaller scale).

pkg/cloud/cloud.go
Jaanki:
```suggestion
type uint16slice []uint16
```

pkg/cloud/cloud.go
```diff
@@ -18,9 +18,54 @@ limitations under the License.
...
+	return fmt.Sprintf("%d", *i)
+}
+
+func (i *uintslice) Set(value string) error {
+	values := strings.Split(value, ",")
+
+	for _, val := range values {
+		tmp, err := strconv.ParseUint(val, 0, 16)
+		if err != nil {
+			return errors.Wrap(err, "Conversion to uint failed")
```
Jaanki:
```suggestion
			return errors.Wrap(err, "Conversion to uint16 failed")
```

skitt (changes requested):
I think defining our own Cobra-influenced type is overkill when `UintSliceVar` works just as well. We’re changing all the port-handling code anyway, so adding `uint16` casts doesn’t bother me as much as adding `uintslice`.

Jaanki (changes requested):

pkg/cloud/cloud.go
maayanf24: I can do the following:
* move the new type declaration `uintslice` to another file 
* use `uintslice` from `cmd\subctl\` 
* use `[]uint16` in `cloud.Ports`

tpantelis: > * move the new type declaration `uintslice` to another file

Please name it `unit16slice`.

skitt: Again, I don’t like having Cobra concerns leak into this API. I would much rather have a `[]uint` here and `uint16()` casts as appropriate elsewhere.

tpantelis: The only potential problem would be losing precision by casting if a user specified a large port number.  But, looking in **cloud-prepare**, the underlying usage of the `PortSpec.Port` field is varied, some convert to `int32`, some `int` and some to a `string`. I think we should just change  `PortSpec.Port` to `uint` and convert appropriately in the cloud provider code. Then we won't need to define a new Cobra type here.

skitt: Ports have to be < 65536 anyway, so if the user specifies a port greater than that, it’s a mistake; we could add a check for that in the Cobra validations. Then as you say, change the ports to `uint`s.

tpantelis: Check out [this commit](https://github.com/submariner-io/subctl/pull/74/commits/4c31ca271c65e483711e09762b322edd70061777). I think the new Cobra type is reasonable  as it validates but I moved it to `cmd/subctl`. The `Ports.Metrics` field is []uint16.

cmd/subctl/uint16slice.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/pkg/errors"
+)
+
+type uint16Slice struct {
+	value *[]uint16
+}
+
+func (s *uint16Slice) Type() string {
+	return "uint16Slice"
```
skitt, line 34: This appears in the help output, to match the `uintslice` output it would be better as `uint16slice`:

```
Flags:
  -h, --help                        help for prepare
      --metrics-port uint16Slice    Metrics port (default [8080 8081])
      --nat-discovery-port uint16   NAT discovery port (default 4490)
      --natt-port uint16            IPSec NAT traversal port (default 4500)
      --vxlan-port uint16           Internal VXLAN port (default 4800)
```

(I suspect that the types in help messages are rather meaningless to many users, but that’s another concern.)

skitt (changes requested):

cmd/subctl/cloud.go
```diff
@@ -58,12 +58,14 @@ func init() {
...
 
+	cloudPorts.Metrics = append(cloudPorts.Metrics, 8080, 8081)
+
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Natt, "natt-port", port.ExternalTunnel,
 		"IPSec NAT traversal port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.NatDiscovery, "nat-discovery-port",
 		port.NATTDiscovery, "NAT discovery port")
 	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Vxlan, "vxlan-port", port.IntraClusterVxLAN, "Internal VXLAN port")
-	cloudPrepareCmd.PersistentFlags().Uint16Var(&cloudPorts.Metrics, "metrics-port", 8080, "Metrics port")
+	cloudPrepareCmd.PersistentFlags().Var(&uint16Slice{value: &cloudPorts.Metrics}, "metrics-port", "Metrics port")
```
skitt: Could we rename this to `metrics-ports` and have `metrics-port` as a deprecated alias?

skitt (changes requested):

pkg/cloud/cloud.go
skitt: Fair enough, thanks!

cmd/subctl/uint16slice.go
tpantelis: So you want the type with a lowercase ‘s’?   I made it upper case to match unitSlice as they have code that looks for “Slice” for certain functionality.

skitt: Ah, no, this is only about the help output. The Cobra `UintSlice` produces `uintslice` somehow in the help output, but I see its `Type()` function returns `uintSlice` so please don’t change that.

tpantelis: yeah I'd think `Type()` is used directly in usage output.

skitt: ... and it’s special-cased here: https://github.com/spf13/pflag/blob/81378bbcd8a1005f72b1e8d7579e5dd7b2d612ab/flag.go#L603 (I got confused with the output from the previous iteration of this PR), producing

```
Flags:
  -h, --help                        help for prepare
      --metrics-port uints          Metrics port (default [8080,8081])
      --nat-discovery-port uint16   NAT discovery port (default 4490)
      --natt-port uint16            IPSec NAT traversal port (default 4500)
      --vxlan-port uint16           Internal VXLAN port (default 4800)
```

cmd/subctl/cloud.go
tpantelis: Done

tpantelis (approved):

skitt (approved):

nyechiel:
@maayanf24 please update the branch and get this merged and backported

PR #78, opened 2022-06-02
diagnose firewall inter-cluster: add support for LB mode

### LINKED ISSUE #1849: subctl diagnose firewall inter-cluster failing when Submariner installed in LB mode
I deployed Submariner in LB mode on aws and gcp, first running the cloud preparation steps [1] and then deployed Submariner using subctl,
although all connectivity tests pass the **subctl diagnostic firewall inter-cluster** reported an error[2].

The 'diagnose firewall inter-cluster' seems to need to be changed to support LB mode as well.

[1]
$ subctl cloud prepare  aws --profile aws-acm-subm  --ocp-metadata install1/metadata.json  --gateways 0 
 ✓ Initializing AWS connectivity
 ✓ Retrieving VPC ID 
 ✓ Retrieved VPC ID vpc-0761008eea28ecb13
 ✓ Validating pre-requisites 
 ✓ Validated pre-requisites
 ✓ Opening port 4800 protocol udp for intra-cluster communications 
 ✓ Opened port 4800 protocol udp for intra-cluster communications
 ✓ Opening port 8080 protocol tcp for intra-cluster communications 
 ✓ Opened port 8080 protocol tcp for intra-cluster communications
 ✓ Opening port 4500 protocol udp for intra-cluster communications 
 ✓ Opened port 4500 protocol udp for intra-cluster communications
 ✓ Opening port 4490 protocol udp for intra-cluster communications 
 ✓ Opened port 4490 protocol udp for intra-cluster communications
 ✓ Opening port 0 protocol 50 for intra-cluster communications 
 ✓ Opened port 0 protocol 50 for intra-cluster communications
 ✓ Opening port 0 protocol 51 for intra-cluster communications 
 ✓ Opened port 0 protocol 51 for intra-cluster communications
$ subctl cloud prepare gcp --ocp-metadata gcp_cluster2/metadata.json  --gateways 0 
 ✓ Retrieving GCP credentials from your GCP configuration
 ✓ Initializing GCP connectivity 
 ✓ Opening internal ports "4800/udp, 8080/tcp" for intra-cluster communications on GCP 
 ✓ Opened internal ports "4800/udp, 8080/tcp" with firewall rule "yblb-7sv57-submariner-internal-ports-ingress" on GCP
$

[2] 
$ subctl diagnose firewall inter-cluster gcp_cluster2/auth/kubeconfig   install1/auth/kubeconfig --verbose 
 ✗ Checking if tunnels can be setup on the gateway node of cluster "clustergcp" 
 ✓ tcpdump output from sniffer pod on Gateway node
 ✓ 
 ✗ The tcpdump output from the sniffer pod does not include the message sent from client pod. Please check that your firewall configuration allows UDP/4500 traffic on the "yblb-7sv57-worker-b-7h5tq.c.gc-acm-dev.internal" node.
$

yboaron:
When I checked the firewall inter-cluster with LB mode I noticed that
both encapsulated and nat discovery traffic received in some platforms
on LB nodeport.

This PR updates the server tcpdump pod to listen also on nodeport in case
SM is deployed in LB mode.

Fixes: https://github.com/submariner-io/submariner/issues/1849

Signed-off-by: yboaron <yboaron@redhat.com>

### DIFF
--- a/pkg/diagnose/firewall.go
+++ b/pkg/diagnose/firewall.go
@@ -45,6 +45,9 @@ const (
 
 const (
 	clientSourcePort = "9898"
+	loadBalancerName = "submariner-gateway"
+	encapsPortName   = "cable-encaps"
+	nattPortName     = "natt-discovery"
 )
 
 const (
@@ -68,6 +71,7 @@ func spawnClientPodOnNonGatewayNode(client kubernetes.Interface, namespace, podC
 	imageRepInfo *image.RepositoryInfo,
 ) (*pods.Scheduled, error) {
 	scheduling := pods.Scheduling{ScheduleOn: pods.NonGatewayNode, Networking: pods.PodNetworking}
+
 	return spawnPod(client, scheduling, "validate-client", namespace, podCommand, imageRepInfo)
 }
 
@@ -220,9 +224,21 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, optio
 		return status.Error(err, "Could not determine the target port")
 	}
 
+	portFilter := fmt.Sprintf("dst port %d", destPort)
+
+	lbNodePort, err := getLbNodePort(localClusterInfo, localEndpoint, targetPort)
+	if err != nil {
+		return status.Error(err, "Could not determine LB node port")
+	}
+
+	// When SM deployed using LB the encapsulated and nat discovery traffic received in some platforms on LB nodeport
+	if lbNodePort != 0 {
+		portFilter += fmt.Sprintf(" or dst port %d ", lbNodePort)
+	}
+
 	clientMessage := string(uuid.NewUUID())[0:8]
-	podCommand := fmt.Sprintf("timeout %d tcpdump -ln -Q in -A -s 100 -i any udp and dst port %d | grep '%s'",
-		options.ValidationTimeout, destPort, clientMessage)
+	podCommand := fmt.Sprintf("timeout %d tcpdump -ln -Q in -A -s 100 -i any udp and %s | grep '%s'",
+		options.ValidationTimeout, portFilter, clientMessage)
 
 	sPod, err := spawnSnifferPodOnNode(localClusterInfo.ClientProducer.ForKubernetes(), gwNodeName, options.PodNamespace, podCommand,
 		localClusterInfo.GetImageRepositoryInfo())
@@ -294,4 +310,30 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
 	default:
 		return 0, fmt.Errorf("could not determine the target port for cable driver %q", endpoint.Spec.Backend)
 	}
+}
+
+func getLbNodePort(clusterInfo *cluster.Info, endpoint *subv1.Endpoint, tgtport TargetPort) (int32, error) {
+	usingLoadBalancer, _ := endpoint.Spec.GetBackendBool(subv1.UsingLoadBalancer, nil)
+	if usingLoadBalancer == nil || !*usingLoadBalancer {
+		return 0, nil
+	}
+
+	portName := encapsPortName
+	if tgtport == NatDiscoveryPort {
+		portName = nattPortName
+	}
+
+	svc, err := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Services(endpoint.GetNamespace()).Get(
+		context.TODO(), loadBalancerName, metav1.GetOptions{})
+	if err == nil {
+		for _, port := range svc.Spec.Ports {
+			if port.Name == portName {
+				return port.NodePort, nil
+			}
+		}
+	} else {
+		return 0, fmt.Errorf("error reading the details of LB service %s: %w", loadBalancerName, err)
+	}
+
+	return 0, fmt.Errorf("could not determine nodePort for port name %q of LB service %s", portName, loadBalancerName)
 }


pkg/diagnose/firewall_tunnel.go
```diff
@@ -87,7 +98,7 @@ func TunnelConfigAcrossClusters(localClusterInfo, remoteClusterInfo *cluster.Inf
 		return false
 	}
 
-	podCommand = fmt.Sprintf("for x in $(seq 1000); do echo %s; done | for i in $(seq 5);"+
+	podCommand = fmt.Sprintf("for x in $(seq 1000); do sleep 0.01 && echo %s; done | for i in $(seq 5);"+
```
sridhargaddam: Is this sleep addressing the issue-B that you mentioned in the note B?

yboaron: Yep, without adding the sleep the 'diagnose inter-cluster' failed on my env

sridhargaddam: > B. I'm not sure if it's specific to LB mode, sometimes the client pod
finished running before the tcpdump server started running.

Ok, Actually in the current code we first spawn the server Pod and after it reaches the Running state, we spawn the client pod. So looks like something is happening with the client pod due to which its getting terminated.

sridhargaddam: Can you please try with HostNetworking on the client Pod

sridhargaddam:
@yboaron I tried running "subctl diagnose firewall inter-cluster ..." on an LB setup and see that its failing. I'll try to investigate the reasons for the failure and will update you.

pkg/diagnose/firewall_tunnel.go
yboaron: I checked the 'diagnose firewall' command  (in LB mode) for both natt and encap ports with client pod runs in host networking  - both tests passed.

I assume that setting the client pod to run in host-networking like the server pod helped with syncing pods creation timing. 

I'm not sure this is the ideal solution for client and server pods synchronization but if it's ok for you it's possible to move forward with this approach.

sridhargaddam: > I assume that setting the client pod to run in host-networking like the server pod helped with syncing pods creation timing

The server pod is always created first and it executes "tcpdump" as soon as it starts running. Here, we modified the client pod from PodNetworking to HostNetworking and this seems to be fixing the issue. Yes, its not clear if HostNetworking is adding a small delay because of which tcpdump in server pod becomes functional and is able to sniff the packets. We have couple of options, either we continue using HostNetwork for clientPods or we can add a small sleep in the client command between the multiple retries is something we can consider.

In parallel, I'm also wondering if we can simplify the serverPod command as shown below.

```
tcpdump -c 10 -lpenni any -Q in udp and dst port <IPSecPort/NatDiscoveryPort>
```

Idea is, we will run the above command on the ServerPod and send traffic from the clientPod (as usual) just to ensure that there is someone sending data to server while the test is going one. If the above tcpdump command returns with some output, that means the respective port is allowed as part of firewall config and the data corresponds to either
1. remote cluster traffic which was already communicating with the Gateway node on that UDP/Port or
2. it could correspond to the data sent from the clientPod to the server pod as part of diagnose validation 

In both the above cases (1 or 2) it implies the firewall allows the UDP/Port on the Gateway node. In this approach we can even reduce the amount of time its taking to run the diagnose firewall commands (as you can see I'm using -c 10 in the tcpdump output)

Note: We configure the firewall rules to allow traffic from any source and do not restrict the sources.
 
What are your thoughts on this?

sridhargaddam: I tried multiple times in my setup and see that HostNetworking seems to be consistently working. I think we can go with HostNetworking for now to unblock the issue and try to optimize it in a separate PR.

pkg/diagnose/firewall.go
```diff
@@ -288,3 +304,32 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
 		return 0, fmt.Errorf("could not determine the target port for cable driver %q", endpoint.Spec.Backend)
 	}
 }
+
+func getLbNodePort(clusterInfo *cluster.Info, endpoint *subv1.Endpoint, tgtport TargetPort, status reporter.Interface) (int32, bool) {
+	var lbNodePort int32
```
tpantelis: This is never set and is only used as an uninitialized return value so we should just remove it and return 0.

pkg/diagnose/firewall.go
```diff
@@ -45,6 +45,9 @@ const (
 
 const (
 	clientSourcePort = "9898"
+	loadBalancerName = "submariner-gateway"
```
tpantelis, line 48: It seems we can use the `names.GatewayComponent` constant rather than defining a new one.

pkg/diagnose/firewall.go
```diff
@@ -288,3 +304,32 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
...
+	svc, err := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Services(endpoint.GetNamespace()).Get(
+		context.TODO(), loadBalancerName, metav1.GetOptions{})
+	if err == nil {
+		for _, port := range svc.Spec.Ports {
+			if port.Name == portName {
+				return port.NodePort, true
+			}
+		}
+	} else {
+		status.Failure(fmt.Sprintf("Error retrieving gateway LB service: %v", err))
```
tpantelis: This is an API failure so we should propagate the error to the caller so it can abort in order to be consistent with the rest of the code base. However we should not treat a `NotFound` error as a failure.

tpantelis (dismissed):

pkg/diagnose/firewall.go
yboaron: Not sure about that, GW LB constants defined [here](https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/loadbalancer_resources.go#L34-#L37) , I think we can either update operator code to expose this constants (preferable option)  or leave it as is.   

Maybe we can update it in a follow up PR after the relevant change on operator is merged. wdyt?

pkg/diagnose/firewall.go
```diff
@@ -220,9 +224,20 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, optio
...
 	}
 
+	portFilter := fmt.Sprintf("dst port %d", destPort)
+
+	lbNodePort, err := getLbNodePort(localClusterInfo, localEndpoint, targetPort)
+	if err != nil {
+		return status.Error(err, "Could not determine LB node port")
+	}
+
+	if lbNodePort != 0 {
```
sridhargaddam, line 235: Please add a note explaining why we are also listening on nodePort for LB mode.

pkg/diagnose/firewall.go
```diff
@@ -45,6 +45,9 @@ const (
 
 const (
 	clientSourcePort = "9898"
+	loadBalancerName = "submariner-gateway"
+	encapsPortName   = "cable-encaps"
```
sridhargaddam, line 49: As this is used only as part of log message, lets use "tunnel port" which is more readable.

pkg/diagnose/firewall.go
```diff
@@ -45,6 +45,9 @@ const (
...
+	nattPortName     = "natt-discovery"
```
sridhargaddam, line 50: Similarly lets use, "Auto NAT discovery port"

pkg/diagnose/firewall.go
```diff
@@ -295,3 +310,29 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
...
+	svc, err := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Services(endpoint.GetNamespace()).Get(
+		context.TODO(), loadBalancerName, metav1.GetOptions{})
+	if err == nil {
+		for _, port := range svc.Spec.Ports {
+			if port.Name == portName {
+				return port.NodePort, nil
+			}
+		}
+	} else {
+		return 0, fmt.Errorf("error reading %s LB: %w", loadBalancerName, err)
```
sridhargaddam:
```suggestion
		return 0, fmt.Errorf("error reading the details of LB service %s: %w", loadBalancerName, err)
```

pkg/diagnose/firewall.go
```diff
@@ -295,3 +310,29 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
...
+	}
+
+	return 0, fmt.Errorf("could not determine LB node port for portname %q", portName)
```
sridhargaddam:
```suggestion
	return 0, fmt.Errorf("could not determine LB node port for %q", portName)
```

sridhargaddam (changes requested):
Thanks for the PR @yboaron.

pkg/diagnose/firewall.go
yboaron: Not sure I understand this comment, these constants are used to retrieve LB port details and not just for logging, port names defined in [operator](https://github.com/submariner-io/submariner-operator/blob/devel/controllers/submariner/loadbalancer_resources.go#L67-#L73)

After we'll merge this PR, I think we should have follow-up PRs, one in the operator (to expose LB constants) and one in subctl to use LB constants from the operator.

pkg/diagnose/firewall.go
yboaron: Same here, not sure I understand your comment

pkg/diagnose/firewall.go
sridhargaddam: Yeah, I see that you are also comparing it in the API getLbNodePort(), its fine then.

sridhargaddam (approved):

skitt (approved):

pkg/diagnose/firewall.go
tpantelis: yeah that sounds fine.

pkg/diagnose/firewall.go
```diff
@@ -295,3 +311,29 @@ func getTargetPort(submariner *v1alpha1.Submariner, endpoint *subv1.Endpoint, tg
...
+		for _, port := range svc.Spec.Ports {
+			if port.Name == portName {
+				return port.NodePort, nil
+			}
+		}
+	} else {
+		return 0, fmt.Errorf("error reading the details of LB service %s: %w", loadBalancerName, err)
+	}
+
+	return 0, fmt.Errorf("could not determine nodePort for port name %q of LB service %s", portName, loadBalancerName)
```
tpantelis, line 338: Is it an error if it's not found?

sridhargaddam: Yes, with LB mode, the Service should exist with nodePort info. If its missing, we should flag it as an error.

yboaron: For the platforms I tested (aws and gcp) nodePort was created for each LB port, I'm pretty sure Nodeport should always be there, so I guess we should report an error if we can not find nodePort.

PR #84, opened 2022-06-06
Bump submariner version

tpantelis:
...using `go get github.com/submariner-io/submariner@devel`. This also updates K8s to 0.23.5. This is to adjust to recent K8s version bumps in submariner due to OVN changes - better to do that now rather than running into problems later in creating the RC0 release.


Jaanki (approved):
A reason for the bump in the commit message would be nice!!

skitt (approved):

PR #88, opened 2022-06-09
Use Submariner CRD image registry info for obtaining nettest image

### LINKED ISSUE #34: `subctl diagnose firewall inter-cluster` references a hard-coded image: `quay.io/submariner/nettest:devel`
The image used for testing the network appears to be hard coded:

`quay.io/submariner/nettest:devel`

I have Submariner `v0.12.0` installed on set of 3 OKD 4.10 clusters.  These clusters are disconnected from the internet.

I am able to successfully install and configure Submariner using the `--repository ${PRIVATE_REGISTRY}/submariner --version ${SUBMARINER_VER}` options.

These options do not appear to be supported in the diagnostic tools.  Thus, the diagnostics fail because the image cannot be pulled.

The relevant code appears to be here:

```bash
./submariner-operator/internal/pods/schedule.go
```

```go
func (np *Scheduled) schedule() error {
	if np.Config.Scheduling.ScheduleOn == CustomNode && np.Config.Scheduling.NodeName == "" {
		return fmt.Errorf("CustomNode is specified for scheduling, but nodeName is missing")
	}

	networkPod := v1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			GenerateName: np.Config.Name,
			Labels: map[string]string{
				"app": np.Config.Name,
			},
		},
		Spec: v1.PodSpec{
			RestartPolicy: v1.RestartPolicyNever,
			HostNetwork:   bool(np.Config.Scheduling.Networking),
			Containers: []v1.Container{
				{
					Name:    np.Config.Name,
					Image:   "quay.io/submariner/nettest:devel",
					Command: []string{"sh", "-c", "$(COMMAND) >/dev/termination-log 2>&1 || exit 0"},
					Env: []v1.EnvVar{
						{Name: "COMMAND", Value: np.Config.Command},
					},
				},
			},
			Tolerations: []v1.Toleration{{Operator: v1.TolerationOpExists}},
		},
	}
```

tpantelis:
...for `subctl diagnose`.

Fixes https://github.com/submariner-io/subctl/issues/34

### DIFF
--- a/internal/pods/schedule.go
+++ b/internal/pods/schedule.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/shipyard/test/e2e/framework"
+	"github.com/submariner-io/subctl/pkg/image"
+	"github.com/submariner-io/submariner-operator/pkg/images"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -52,12 +54,13 @@ type Scheduling struct {
 }
 
 type Config struct {
-	Name       string
-	ClientSet  kubernetes.Interface
-	Scheduling Scheduling
-	Namespace  string
-	Command    string
-	Timeout    uint
+	Name                string
+	ClientSet           kubernetes.Interface
+	Scheduling          Scheduling
+	Namespace           string
+	Command             string
+	Timeout             uint
+	ImageRepositoryInfo image.RepositoryInfo
 }
 
 type Scheduled struct {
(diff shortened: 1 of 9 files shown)


internal/pods/schedule.go
```diff
@@ -124,7 +130,7 @@ func (np *Scheduled) schedule() error {
 			Containers: []v1.Container{
 				{
 					Name:    np.Config.Name,
-					Image:   "quay.io/submariner/nettest:devel",
+					Image:   fmt.Sprintf("%s/nettest:devel", imageRepo),
```
skitt: I think this should also use the right version tag, for a variety of reasons; at least (a) we can’t expect `nettest:devel` to always work for us (on a release branch); and (b) we can’t expect users to pull a `:devel` image when they prepare a set of images for disconnected provisioning, instead of whatever version they’re setting up. See `images.GetImagePath()`, I think that would be useful here.

skitt (changes requested):

internal/pods/schedule.go
tpantelis: ok - makes sense. I changed it to extract the image registry info from the `Submariner` CRD into an `image.RepositoryInfo` struct and use `images.GetImagePath(...)`.

skitt (approved):

Jaanki (approved):

PR #97, opened 2022-06-14
Use binaries from cmd/bin

skitt:
We build subctl in cmd/bin, add that to the PATH instead of bin to
ensure we test the binary we're building.


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
> [E2E (globalnet, 1.23, lighthouse)](https://github.com/submariner-io/subctl/runs/6885442328?check_suite_focus=true#logs)

Timed out, re-running

PR #99, opened 2022-06-15
Add OVNKubernetes diagnosis

### LINKED ISSUE #92: Add OVN diagnostics to diagnose
Check the following when running diagnose on clusters using OVNKubernetes as CNI

1. Detect if OVNK is the CNI
2. Check if `submariner-networkplugin-syncer` pod is running correctly.
3. Check if the OVN version is compatible with submariner.
4. ~~Check if all OVN resources were added correctly.~~ We don't do that for any CNI. Will be added later if needed.

**Why is this needed**:
Currently diagnose doesn't check all OVN relevant configuration in a deployment, so can't detect any OVN specific errors.

vthapar:
* Check status of networkplugin-syncer deployment
* Check version of OVN NBDB

Fixes #92

### DIFF
--- a/pkg/diagnose/cni.go
+++ b/pkg/diagnose/cni.go
@@ -237,16 +237,16 @@ func checkOVNVersion(info *cluster.Info, status reporter.Interface) bool {
 
 	ovnNBVersion, err := getOVNNBVersion(clientSet, info.RestConfig, ovnPod)
 	if err != nil {
-		status.Failure("Failed to get OVN NB version %v", err)
+		status.Failure("Failed to get ovn-nb database version %v", err)
 		return false
 	}
 
 	if version.Compare(ovnNBVersion, minOVNNBVersion, "<") {
-		status.Failure("The OVN NB DB version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
+		status.Failure("The ovn-nb database version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
 		return false
 	}
 
-	status.Success("The OVN NB DB version %v is supported", ovnNBVersion)
+	status.Success("The ovn-nb database version %v is supported", ovnNBVersion)
 
 	return true
 }
@@ -275,6 +275,7 @@ func getOVNNBVersion(clientSet kubernetes.Interface, config *rest.Config, pod *c
 		// NBDB container name is nb-ovsdb [vanilla OVNK] or nbdb [OCP].
 		if strings.HasPrefix(container.Name, "nb") {
 			containerName = container.Name
+			break
 		}
 	}
 
--- a/pkg/diagnose/cni.go
+++ b/pkg/diagnose/cni.go
@@ -19,18 +19,30 @@ limitations under the License.
 package diagnose
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/mcuadros/go-version"
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/subctl/pkg/cluster"
 	submv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cni"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	ovnKubeDBPodLabel = "ovn-db-pod=true"
+	minOVNNBVersion   = "6.1.0"
 )
 
 var supportedNetworkPlugins = []string{
@@ -70,6 +82,10 @@ func CNIConfig(clusterInfo *cluster.Info, status reporter.Interface) bool {
 			" It may or may not work.", clusterInfo.Submariner.Status.NetworkPlugin)
 	} else {
 		status.Success("The detected CNI network plugin (%q) is supported", clusterInfo.Submariner.Status.NetworkPlugin)
+	}
+
+	if clusterInfo.Submariner.Status.NetworkPlugin == cni.OVNKubernetes {
+		return checkOVNVersion(clusterInfo, status)
 	}
 
 	return checkCalicoIPPoolsIfCalicoCNI(clusterInfo, status)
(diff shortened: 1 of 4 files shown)


pkg/diagnose/cni.go
```diff
@@ -206,3 +222,93 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	clientSet := info.ClientProducer.ForKubernetes()
+
+	ovnPod, err := findPod(clientSet, ovnKubeDBPodLabel)
+	if err != nil || ovnPod == nil {
+		status.Failure("Failed to get OVNKubeDB Pod %v", err)
+		return false
+	}
+
+	ovnNBVersion, err := getOVNNBVersion(clientSet, info.RestConfig, ovnPod)
+	if err != nil {
```
astoycos, line 239: Duplicate Error Message

pkg/diagnose/cni.go
```diff
@@ -19,18 +19,30 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+const (
+	ovnKubeDBPodLabel = "ovn-db-pod=true"
```
astoycos, line 44: I'm not sure if we use the same labels downstream since all ovn-k containers are in a pod

astoycos: Oh maybe it will
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml#L54

Just be cautious in the OCP deployment nbdb sbdb are just containers running in the same pod

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,93 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+
+	if len(pods.Items) == 0 {
+		return nil, errors.WithMessagef(err, "found 0 pods with label %v", labelSelector)
+	}
+
+	return &pods.Items[0], nil
+}
+
+func getOVNNBVersion(clientSet kubernetes.Interface, config *rest.Config, pod *corev1.Pod) (string, error) {
+	// We can run ovn-nbctl in any of the containers in db pod
```
astoycos: No DB pod downstream, I'd be sure to get in the nbdb container to be safe

astoycos:
I don't love this but it'll work  :) 

Most of this info is also probably in OCP MustGathers as well

pkg/diagnose/cni.go
vthapar: Problem is the name of nbdb container varies between vanilla OVNK [used in kind] and OCP. Not sure about downstream, but I tested with upstream version. These are the containers in `ovnkube-master` pod

`northd, nbdb, kube-rbac-proxy, sbdb, ovnkube-master, ovn-dbchecker`

`ovn-nbctl` works in all except for `kube-rbac-proxy`. A cleaner way would be to go through container names and pick the nbdb one.

EDIT: Modified code to look for nb container than pick `[0]`

pkg/diagnose/cni.go
vthapar: Yeah, saw that. What matters for us is if it can run ovn-nbctl and that works in both containers.

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	status.Start("Checking OVN version")
+	defer status.End()
+
+	clientSet := info.ClientProducer.ForKubernetes()
+
+	ovnPod, err := findPod(clientSet, ovnKubeDBPodLabel)
+	if err != nil || ovnPod == nil {
+		status.Failure("Failed to get OVNKubeDB Pod %v", err)
+		return false
+	}
```
tpantelis: `findPod` returns an error if not found so no need to check for nil `ovnPod`.

```suggestion
	if err != nil {
		status.Failure("Failed to get the OVN DB pod: %v", err)
		return false
	}
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+
+	ovnNBVersion, err := getOVNNBVersion(clientSet, info.RestConfig, ovnPod)
+	if err != nil {
+		status.Failure("Failed to get OVN NB version %v", err)
+		return false
+	}
+
+	if version.Compare(ovnNBVersion, minOVNNBVersion, "<") {
+		status.Failure("OVN NBDB version %v is lesser than minimum supported %v", ovnNBVersion, minOVNNBVersion)
```
tpantelis:
```suggestion
		status.Failure("The OVN NB DB version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+		return false
+	}
+
+	status.Success("OVN NBDB Version %v is supported", ovnNBVersion)
```
tpantelis:
```suggestion
	status.Success("The OVN NB DB version %v is supported", ovnNBVersion)
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+
+	return true
+}
+
+func findPod(clientSet kubernetes.Interface, labelSelector string) (*corev1.Pod, error) {
```
tpantelis:
```suggestion
func mustFindPod(clientSet kubernetes.Interface, labelSelector string) (*corev1.Pod, error) {
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	pods, err := clientSet.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labelSelector,
+		Limit:         1,
+	})
+	if err != nil {
+		return nil, errors.WithMessagef(err, "error listing Pods by label selector %q", labelSelector)
+	}
+
+	if len(pods.Items) == 0 {
+		return nil, errors.WithMessagef(err, "found 0 pods with label %v", labelSelector)
```
tpantelis:
```suggestion
		return nil, errors.WithMessagef(err, "no pod found with label %v", labelSelector)
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+		Stdout:    true,
+		Stderr:    true,
+		TTY:       false,
+	}, scheme.ParameterCodec)
+
+	var stdout, stderr bytes.Buffer
+
+	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
+	if err != nil {
+		return "", errors.WithMessagef(err, "Failed to get OVN NBDB version")
```
tpantelis:
```suggestion
		return "", errors.WithMessagef(err, "failed to create SPDY executor")
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+
+	err = exec.Stream(remotecommand.StreamOptions{
+		Stdin:  nil,
+		Stdout: &stdout,
+		Stderr: &stderr,
+		Tty:    false,
+	})
+
+	if err != nil {
+		return "", errors.WithMessagef(err, "Failed to get OVN NBDB version")
```
tpantelis:
```suggestion
		return "", errors.WithMessagef(err, "failed to execute SPDY command")
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,97 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	}
+
+	result := strings.Split(stdout.String(), "DB Schema")[1]
```
tpantelis: We should check that the split yielded 2 elements to avoid a crash.

tpantelis (changes requested):

pkg/diagnose/cni.go
vthapar: This is a user facing msg, so went with something that makes sense for user. Thoughts?

tpantelis: yeah but the caller will wrap it with the same message. Nested messages get more detailed and technical. When troubleshooting a failure we'll need the details and the message I proposed here also gives the location of the call site.

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,101 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	})
+
+	if err != nil {
+		return "", errors.WithMessagef(err, "failed to execute SPDY command")
+	}
+
+	results := strings.Split(stdout.String(), "DB Schema")
+
+	if len(results) < 2 {
+		return "", errors.WithMessagef(err, "OVN NB DB Schema version not found")
```
tpantelis: Minor thing but if this occurred it would be useful to see the output.

```suggestion
		return "", errors.WithMessagef(err, "unable to determine the version from the ovn-nbctl output: %q", stdout.String())
```

tpantelis (changes requested):

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,101 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+}
+
+func getOVNNBVersion(clientSet kubernetes.Interface, config *rest.Config, pod *corev1.Pod) (string, error) {
+	containerName := ""
+
+	for i := 0; i < len(pod.Spec.Containers); i++ {
+		container := pod.Spec.Containers[i]
+		// NBDB container name is nb-ovsdb [vanilla OVNK] or nbdb [OCP].
+		if strings.HasPrefix(container.Name, "nb") {
+			containerName = container.Name
```
sridhargaddam, line 277: break?

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,101 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+	}
+
+	ovnNBVersion, err := getOVNNBVersion(clientSet, info.RestConfig, ovnPod)
+	if err != nil {
+		status.Failure("Failed to get OVN NB version %v", err)
+		return false
+	}
+
+	if version.Compare(ovnNBVersion, minOVNNBVersion, "<") {
+		status.Failure("The OVN NB DB version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
```
sridhargaddam:
```suggestion
		status.Failure("The ovn-nb database version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
```

Looking at the man pages, the convention seems to be ovn-nb database.

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,101 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+		return false
+	}
+
+	status.Success("The OVN NB DB version %v is supported", ovnNBVersion)
```
sridhargaddam:
```suggestion
	status.Success("The ovn-nb database version %v is supported", ovnNBVersion)
```

pkg/diagnose/cni.go
```diff
@@ -206,3 +222,101 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
+
+	ovnPod, err := mustFindPod(clientSet, ovnKubeDBPodLabel)
+	if err != nil {
+		status.Failure("Failed to get OVNKubeDB Pod %v", err)
+		return false
+	}
+
+	ovnNBVersion, err := getOVNNBVersion(clientSet, info.RestConfig, ovnPod)
+	if err != nil {
+		status.Failure("Failed to get OVN NB version %v", err)
```
sridhargaddam:
```suggestion
		status.Failure("Failed to get ovn-nb database version %v", err)
```

sridhargaddam (changes requested):
Some minor comments, otherwise LGTM. Thanks for the PR @vthapar

Jaanki (approved):

sridhargaddam (approved):

aswinsuryan (approved):

sridhargaddam:
@skitt @mkolesnik e2e is failing with the following error

```
[06:44:31.598] [subctl]$ [cluster2] find /go/src/github.com/submariner-io/subctl/output/kubeconfigs -type f -printf %p:
[06:44:31.606] [subctl]$ [cluster2] bash -c curl -Ls https://get.submariner.io/ | VERSION=devel DESTDIR=/go/bin bash
bash: line 1: html: No such file or directory
bash: line 2: syntax error near unexpected token `<'
bash: line 2: `<head><title>500 Internal Server Error</title></head>
'
[06:44:47.224] [subctl]$ [cluster2] bash -c curl -Ls https://get.submariner.io/ | VERSION=devel DESTDIR=/go/bin bash
bash: line 1: html: No such file or directory
bash: line 2: syntax error near unexpected token `<'
bash: line 2: `<head><title>500 Internal Server Error</title></head>
'
make: *** [/opt/shipyard/Makefile.inc:134: deploy] Error 2
```

sridhargaddam:
E2e passed after recheck

sridhargaddam:
@tpantelis can you please take a re-look.

tpantelis (approved):

PR #101, opened 2022-06-16
Cloud prepare support for OVN Clusters

### LINKED ISSUE #90: OVN support in cloud prepare
Only open VxLan ports required if OVNKubernetes is the used CNI. This requires:

1. Detect if OVNKubernetes is the CNI
2. If yes, open the minimal set of ports.

**Why is this needed**:
When using OVNKubernetes as CNI we don't need to open all the vxlan ports in the cloud. So only open the vxlan ports required based on which CNI is used.

sridhargaddam:
When deploying Submariner on an OCP Cluster with OVNK as
the CNI, Submariner does not use the internal VxLAN tunnels
to forward the traffic from the worker nodes to the local
Gateway node. So as part of cloud prepare, we can avoid
opening of this port. This PR handles this for AWS, GCP,
OSP and Azure clusters.

Fixes: https://github.com/submariner-io/subctl/issues/90
Signed-off-by: Sridhar Gaddam <sgaddam@redhat.com>

### DIFF
--- a/pkg/cloud/prepare/prepare.go
+++ b/pkg/cloud/prepare/prepare.go
@@ -1,0 +1,100 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package prepare
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/cloud-prepare/pkg/api"
+	"github.com/submariner-io/subctl/internal/constants"
+	"github.com/submariner-io/subctl/internal/restconfig"
+	"github.com/submariner-io/subctl/pkg/cloud"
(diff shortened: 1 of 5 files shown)


Jaanki (approved):

skitt (approved):

pkg/cloud/prepare/prepare.go
```diff
@@ -0,0 +1,100 @@
...
+	if useNumericESP {
+		for i, port := range gwPorts {
+			switch port.Protocol {
+			case "esp":
+				gwPorts[i].Protocol = "50"
+			case "ah":
+				gwPorts[i].Protocol = "51"
+			}
+		}
+	}
```
tpantelis, line 76: How about we pass the additional ports instead of a kind of esoteric boolean flag:

```Go
func getPortConfig(restcfg *restconfig.Producer, ports *cloud.Ports, additionalPorts ...api.PortSpec,
) ([]api.PortSpec, api.PrepareForSubmarinerInput, error) {
	gwPorts := []api.PortSpec{
		{Port: ports.Natt, Protocol: "udp"},
		{Port: ports.NatDiscovery, Protocol: "udp"},
	}

         gwPorts = append(gwPorts, additionalPorts...)

          ...
}
```

tpantelis (changes requested):

pkg/cloud/prepare/prepare.go
sridhargaddam: The purpose of this API is to return the list of ports to be opened for the clusters and the reason for passing the boolean flag is as follows.

ESP/AH is the name of protocol just like TCP/UDP. Each protocol also has an equivalent numeric value. 
Unfortunately, AWS alone does not accept "esp/ah" as protocol and requires us to configure them as numeric values. 

I'd prefer to keep it in the current way as its more readable rather than passing additional ports. I hope you are fine with it as its just a matter of choice/preferance and not any real issue as such.

tpantelis (approved):

PR #104, opened 2022-06-16
Set client pod of firewall inter-cluster to run in hostnet

yboaron:
We have noticed that with the latest OCP deployments the
diagnose inter-cluster client pod is sometimes getting terminated
due to which the validation fails, using HostNetworking
for the clientPod works consistently.

This PR updates client pod used in firewal inter-cluster command to
run in HostNetworking instead of PodNetworking.

Signed-off-by: yboaron <yboaron@redhat.com>

### DIFF
--- a/pkg/diagnose/firewall.go
+++ b/pkg/diagnose/firewall.go
@@ -71,6 +71,13 @@ func spawnClientPodOnNonGatewayNode(client kubernetes.Interface, namespace, podC
 	return spawnPod(client, scheduling, "validate-client", namespace, podCommand, imageRepInfo)
 }
 
+func spawnClientPodOnNonGatewayNodeWithHostNet(client kubernetes.Interface, namespace, podCommand string,
+	imageRepInfo *image.RepositoryInfo,
+) (*pods.Scheduled, error) {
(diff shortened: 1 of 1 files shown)


pkg/diagnose/firewall.go
```diff
@@ -65,9 +65,15 @@ func spawnSnifferPodOnGatewayNode(client kubernetes.Interface, namespace, podCom
 }
 
 func spawnClientPodOnNonGatewayNode(client kubernetes.Interface, namespace, podCommand string,
-	imageRepInfo *image.RepositoryInfo,
+	imageRepInfo *image.RepositoryInfo, isHostNet bool,
```
sridhargaddam: Lets create a separate API for HostNetworking, similar to the way we are doing in the release-0.12 branch - 
https://github.com/submariner-io/submariner-operator/blob/39f03e60e4b70b15fb2dec88321f1b31abab3805/pkg/subctl/cmd/diagnose/firewall.go#L241

sridhargaddam (changes requested):

sridhargaddam (approved):
Thanks for pushing the PR @yboaron

tpantelis (approved):

PR #107, opened 2022-06-20
Allow broker certificate checks to be disabled

### LINKED ISSUE #52: Add support for BrokerK8sInsecure to subctl
Submariner can disable certificate verification when contacting the broker; this is configured using `BrokerK8sInsecure` in the `Submariner` object. `subctl` should allow enabling this, and apply the setting to itself when contacting the broker.

skitt:
This allows connections to the broker without verifying the broker TLS
certificate, which can be useful for example if the certificates can't
be shared. The option is flipped compared to the Submariner CR field,
but this matches our policy for boolean flags.

To test this, deploy the broker as usual, then empty the contents of
"ca.crt" in the broker-info.subm file; a regular "subctl join" will
fail, but "subctl join --check-broker-certificate=false" will succeed
(and the connected clusters will connect to the broker without
verifying the TLS certificate).

Fixes: #52


dfarrell07 (approved):

tpantelis (approved):

nyechiel:
(adding to the 0.13 board for visibility; we need to mention this change and new the `subctl` flag in the release notes)

skitt:
Note that this was merged after 0.13.0 so shouldn’t be part of the 0.13.0 release notes.

PR #111, opened 2022-06-21
Separate system tests out of E2E

mkolesnik:
This PR separates the system tests out of E2E, making E2E super simple and system tests more focused.

As part of this work the E2E tests will be the default ones (as run by Shipyard).
The system tests have been improved a lot as part of this effort.

Please see individual commits for more details.


.github/workflows/system.yml
```diff
@@ -1,22 +1,27 @@
...
+name: System Tests
 
 on:
   pull_request:
 
 jobs:
-  test-subctl:
-    name: subctl kubeconfig commands
+  system:
+    name: Functionality
```
tpantelis:
```suggestion
    name: Execute subctl commands
```

tpantelis (changes requested):

tpantelis (approved):

Makefile
```diff
@@ -75,8 +76,8 @@ export PATH := $(CURDIR)/cmd/bin:$(PATH)
 # (with the PATH set above)
 deploy: cmd/bin/subctl
 
-e2e: deploy
-	scripts/kind-e2e/e2e.sh $(E2E_ARGS)
+system: deploy
```
skitt: `system` on its own doesn’t convey the meaning of the target; I think `test`, or `system-test`, would be clearer. Since the goal is to test `subctl` (as illustrated by subsequent commits), `test` feels appropriate.

skitt (changes requested):

Makefile
skitt: I don’t see the fix…

Makefile
```diff
@@ -75,8 +76,8 @@ export PATH := $(CURDIR)/cmd/bin:$(PATH)
 # (with the PATH set above)
 deploy: cmd/bin/subctl
 
-e2e: deploy
-	scripts/kind-e2e/e2e.sh $(E2E_ARGS)
+system:
```
skitt: This still says `system`.

skitt (changes requested):

skitt (approved):

scripts/kind-e2e/lib_operator_verify_subm.sh
```diff
@@ -1,491 +0,0 @@
...
-function verify_subm_gateway_pod() {
-  kubectl wait --for=condition=Ready pods -l app=$gateway_deployment_name --timeout=120s --namespace=$subm_ns
-
-  subm_gateway_pod_name=$(kubectl get pods --namespace=$subm_ns -l app=$gateway_deployment_name -o=jsonpath='{.items..metadata.name}')
-
-  json_file=/tmp/${subm_gateway_pod_name}.${cluster}.json
-  kubectl get pod $subm_gateway_pod_name --namespace=$subm_ns -o json | tee $json_file
-
-  validate_pod_container_equals 'image' "${subm_gateway_image_repo}/submariner-gateway:local"
-  validate_pod_container_has 'securityContext.capabilities.add' 'net_admin'
```
dfarrell07, line 262: Am I right in understanding that the new tests don't actually assert details at this level? They just check for the presence of K8s objects?

mkolesnik: Yes as this knowledge is internal to operator and tested there, here its not necessary nor helpful to have these white box tests as the operator takes care of these details.

dfarrell07: Ah so we're keeping these tests in https://github.com/submariner-io/submariner-operator/pull/2104 but removing them here, I see and that makes sense.

PR #121, opened 2022-06-22
show brokers command to show broker spec

### LINKED ISSUE #115: `subctl show brokers` not actually showing any useful information
**What happened**:
I run `subctl show brokers` and get nothing useful.
```sh
# subctl show brokers
Cluster "cluster1"
 ✓ Detecting broker(s)
NAMESPACE                NAME                     COMPONENTS                              
submariner-k8s-broker    submariner-broker        service-discovery, connectivity         

Cluster "cluster2"
 ✓ Detecting broker(s)
```

At least the spec from the defined broker should be shown (perhaps not in yaml), ie:
```yaml
Spec:
  Components:
    service-discovery
    connectivity
  Default Globalnet Cluster Size:  65536
  Globalnet CIDR Range:            242.0.0.0/8
  Globalnet Enabled:               true
```

Im not sure if this is a bug or an enhancement :)
I think we could regard this as a bug, since `subctl show brokers` should give some useful info besides the fact where broker is installed.

FWIW I don't think extended info should be shown in `subctl show all` but we could debate that.

**Environment**:
- Install tools:
- Others:

Jaanki:
`subctl show brokers` commands will now show the broker spec
along with the existing information. Below are the new fields added:

* DefaultCustomDomains
* GlobalnetCIDRRange
* DefaultGlobalnetClusterSize
* GlobalnetEnabled

The output would look like this

```Cluster "cluster1"
 ✓ Detecting broker(s)
NAMESPACE                   NAME                     COMPONENTS                            GLOBALNET ENABLED       GLOBALNET CIDR        DEFAULT GLOBALNET SIZE        DEFAULT DOMAINS                                             
submariner-k8s-broker       submariner-broker        connectivity                          false                   242.0.0.0/8           65536                                                                                     

```

Closes: #115


nyechiel:
@Jaanki do you happen to have an example output with this change?

internal/show/brokers.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
-	template := "%-25.24s%-25.24s%-40.39s\n"
+	template := "%-14.28s%-14.28s%-14.28s%-14.28s%-14.28s%-14.28s%-14.28s\n"
```
Jaanki: This column width needs to be reconfirmed once I have a running cluster.

internal/show/brokers.go
```diff
@@ -51,14 +51,19 @@ func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
...
+	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS", "DEFAULT CUSTOM DOMAINS", "DEFAULT GLOBALNET CLUSTER SIZE",
+		"GLOBALNET CIDR RANGE", "GLOBALNET ENABLED")
 
 	for i := range brokers {
 		fmt.Printf(
 			template,
 			brokers[i].Namespace,
 			brokers[i].Name,
 			strings.Join(brokers[i].Spec.Components, ", "),
+			brokers[i].Spec.DefaultCustomDomains,
```
tpantelis: This is an array as well.

```suggestion
			strings.Join(brokers[i].Spec.DefaultCustomDomains, ", "),
```

tpantelis (changes requested):

Jaanki:
> @Jaanki do you happen to have an example output with this change?

@nyechiel Updated the PR description with the output.

internal/show/brokers.go
```diff
@@ -51,14 +51,19 @@ func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
 		return true
 	}
 
-	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS")
+	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS", "DEFAULT GLOBALNET CLUSTER SIZE", "GLOBALNET CIDR RANGE", "GLOBALNET ENABLED",
+		"DEFAULT CUSTOM DOMAINS")
```
tpantelis: I suggest we put the more useful columns first.

```suggestion
	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS", "GLOBALNET ENABLED", "GLOBALNET CIDR RANGE", "DEFAULT GLOBALNET CLUSTER SIZE", "DEFAULT CUSTOM DOMAINS")
```

tpantelis (dismissed):

internal/show/brokers.go
mkolesnik:
```suggestion
	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS", "GLOBALNET", "GLOBALNET CIDR", "DEFAULT GLOBALNET SIZE", "DEFAULT DOMAINS")
```

I suggest we also drop some superfluous terms from the header to make it more compact (and update the format accordingly).

Jaanki: > I suggest we also drop some superfluous terms from the header to make it more compact (and update the format accordingly).

These are very confusing for me to understand. But I don't have a strong opinion about this. So I don't mind changing except `GLOBALNET ENABLED` because its value is bool. And showing `GLOBALNET ENABLED True` makes more sense.

mkolesnik: We could easily show yes/no instead of true/false which is also more user friendly but OK

internal/show/brokers.go
```diff
@@ -51,14 +49,20 @@ func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
 		return true
 	}
 
-	fmt.Printf(template, "NAMESPACE", "NAME", "COMPONENTS")
+	template := "%-28s%-25.24s%-38s%-24v%-22s%-30v%-60s\n"
```
mkolesnik:
```suggestion
	template := "%-25.24s%-25.24s%-34.33s%-20v%-21s%-26v%-40s\n"
```

I'd use a more compact scheme, resulting in 192 character long print (instead of 228):
```
NAMESPACE                NAME                     COMPONENTS                        GLOBALNET ENABLED   GLOBALNET CIDR       DEFAULT GLOBALNET SIZE    DEFAULT DOMAINS                         
submariner-k8s-broker    submariner-broker        connectivity, service-discovery   false               242.0.0.0/8          65536                                                             
```

mkolesnik (approved):

skitt (approved):

sridhargaddam (approved):

dfarrell07:
PR scrub agrees this is good to backport to 0.13.

PR #125, opened 2022-06-23
Create secrets for serviceaccounts

### LINKED ISSUE #102: In K8s 1.22/1.24 and later, SAs don’t automatically get a secret with a token
`subctl` needs to create a number of service accounts, and relies on Kubernetes to create an associated secret and token. Starting with Kubernetes 1.22, secret/token generation is no longer guaranteed, and it is disabled by default starting with Kubernetes 1.24.

`subctl` needs to be fixed to create a secret after creating a service account, and wait for the token on the secret if necessary. See https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets for details.

Without this, Submariner can no longer be installed on Kubernetes 1.24 or later.

The fix will have to be backported to 0.12.

vthapar:
K8s 1.24+ doesn't autocreate secret for serviceaccount. This
breaks submariner deployments which expects secret to be created
and associated to the corresponding serviceaccount.

The fix ensures the Secret of type service-account-token exists,
is annotated with ServiceAccount.Name, and is added to Secrets list in
the ServiceAccount.

Fixes #102

### DIFF
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -91,7 +91,7 @@ func Ensure(crdUpdater crd.Updater, kubeClient kubernetes.Interface, componentAr
 
 func createBrokerClusterRoleAndDefaultSA(kubeClient kubernetes.Interface, inNamespace string) error {
 	// Create the a default SA for cluster access (backwards compatibility with documentation)
-	_, err := CreateNewBrokerSA(kubeClient, submarinerBrokerClusterDefaultSA, inNamespace)
+	err := CreateNewBrokerSA(kubeClient, submarinerBrokerClusterDefaultSA, inNamespace)
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return errors.Wrap(err, "error creating the default broker service account")
 	}
@@ -115,7 +115,7 @@ func createBrokerClusterRoleAndDefaultSA(kubeClient kubernetes.Interface, inName
 func CreateSAForCluster(kubeClient kubernetes.Interface, clusterID, inNamespace string) (*v1.Secret, error) {
 	saName := names.ForClusterSA(clusterID)
 
-	_, err := CreateNewBrokerSA(kubeClient, saName, inNamespace)
+	err := CreateNewBrokerSA(kubeClient, saName, inNamespace)
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return nil, errors.Wrap(err, "error creating cluster sa")
 	}
@@ -135,7 +135,7 @@ func CreateSAForCluster(kubeClient kubernetes.Interface, clusterID, inNamespace 
 
 func createBrokerAdministratorRoleAndSA(kubeClient kubernetes.Interface, inNamespace string) error {
 	// Create the SA we need for the managing the broker (from subctl, etc..).
-	_, err := CreateNewBrokerSA(kubeClient, constants.SubmarinerBrokerAdminSA, inNamespace)
+	err := CreateNewBrokerSA(kubeClient, constants.SubmarinerBrokerAdminSA, inNamespace)
 	if err != nil && !apierrors.IsAlreadyExists(err) {
 		return errors.Wrap(err, "error creating the broker admin service account")
 	}
@@ -204,19 +204,9 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
 }
 
 // nolint:wrapcheck // No need to wrap here
-func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (brokerSA *v1.ServiceAccount, err error) {
+func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (err error) {
 	sa := NewBrokerSA(submarinerBrokerSA)
 	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa, true)
 
-	if err != nil {
-		return nil, err
-	}
-
-	_, err = serviceaccount.EnsureSecretFromSA(kubeClient, sa.Name, inNamespace)
-
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get secret for broker SA")
-	}
-
-	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Get(context.TODO(), sa.Name, metav1.GetOptions{})
+	return err
 }
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -30,6 +30,7 @@ import (
 	"github.com/submariner-io/subctl/pkg/gateway"
 	"github.com/submariner-io/subctl/pkg/namespace"
 	"github.com/submariner-io/subctl/pkg/role"
+	"github.com/submariner-io/subctl/pkg/serviceaccount"
 	"github.com/submariner-io/submariner-operator/pkg/crd"
 	"github.com/submariner-io/submariner-operator/pkg/lighthouse"
 	"github.com/submariner-io/submariner-operator/pkg/names"
@@ -204,6 +205,18 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
 
 // nolint:wrapcheck // No need to wrap here
 func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (brokerSA *v1.ServiceAccount, err error) {
-	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Create(
-		context.TODO(), NewBrokerSA(submarinerBrokerSA), metav1.CreateOptions{})
+	sa := NewBrokerSA(submarinerBrokerSA)
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = serviceaccount.EnsureSecretFromSA(kubeClient, sa.Name, inNamespace)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get secret for broker SA")
+	}
+
+	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Get(context.TODO(), sa.Name, metav1.GetOptions{})
 }
--- a/pkg/broker/ensure.go
+++ b/pkg/broker/ensure.go
@@ -206,7 +206,7 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
 // nolint:wrapcheck // No need to wrap here
 func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (brokerSA *v1.ServiceAccount, err error) {
 	sa := NewBrokerSA(submarinerBrokerSA)
-	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa)
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa, true)
 
 	if err != nil {
 		return nil, err
--- a/pkg/serviceaccount/ensure.go
+++ b/pkg/serviceaccount/ensure.go
@@ -53,7 +53,7 @@ func ensureFromYAML(kubeClient kubernetes.Interface, namespace, yaml string) (*c
 		return nil, err
 	}
 
-	_, err = Ensure(kubeClient, namespace, sa, true)
+	err = ensure(kubeClient, namespace, sa, true)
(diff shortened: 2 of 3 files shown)


pkg/secret/ensure.go
```diff
@@ -46,3 +60,85 @@ func Ensure(client kubernetes.Interface, namespace string, secret *v1.Secret) (*
...
+		Name: saSecret.Name,
+	}
+
+	sa.Secrets = append(sa.Secrets, secretRef)
+	_, err = serviceaccount.Ensure(client, namespace, sa)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to update ServiceAccount %v with Secret reference %v", saName, secretRef.Name)
+	}
+
+	return saSecret, err
```
tpantelis:
```suggestion
	return saSecret, nil
```

pkg/secret/ensure.go
```diff
@@ -46,3 +60,85 @@ func Ensure(client kubernetes.Interface, namespace string, secret *v1.Secret) (*
...
+	for _, saSecretRef := range sa.Secrets {
+		if strings.HasPrefix(saSecretRef.Name, secretNamePrefix) {
+			saSecret, _ := client.CoreV1().Secrets(sa.Namespace).Get(context.TODO(), saSecretRef.Name, metav1.GetOptions{})
+			if saSecret.Annotations[saAnnotation] == sa.Name && saSecret.Type == v1.SecretTypeServiceAccountToken {
+				return saSecret, nil
+			}
+		}
+	}
+
+	return nil, errors.New(fmt.Sprintf("secret token for service-account %v/%v not found", sa.Namespace, sa.Name))
```
tpantelis: We don't need to return an error if not found - simply return nil `*v1.Secret` . There's no actual error that can occur so remove it as a return value.

pkg/secret/ensure.go
```diff
@@ -46,3 +60,85 @@ func Ensure(client kubernetes.Interface, namespace string, secret *v1.Secret) (*
...
+		return nil, err
+	}
+
+	for _, saSecret := range saSecrets.Items {
+		if saSecret.Annotations[saAnnotation] == sa.Name {
+			return &saSecret, nil
+		}
+	}
+
+	return nil, errors.New(fmt.Sprintf("secret for service-account %v/%v not found", sa.Namespace, sa.Name))
```
tpantelis: We should return `NotFound` error so the caller can distinguish it.

```suggestion
	return nil, apierrors.NewNotFound(schema.GroupResource{
	    Group:    v1.SchemeGroupVersion.Group,
             Resource: "secrets",
        }, sa.Name)
```

pkg/broker/ensure.go
```diff
@@ -204,6 +206,18 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
...
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = secret.EnsureFromSA(kubeClient, sa.Name, inNamespace)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to get secret for broker SA")
```
tpantelis:
```suggestion
		return nil, errors.Wrap(err, "failed to get secret for broker SA")
```

pkg/diagnose/k8s_version.go
```diff
@@ -44,7 +46,13 @@ func K8sVersion(client kubernetes.Interface, status reporter.Interface) bool {
 		return false
 	}
 
-	status.Success("Kubernetes version %q is supported", k8sVersion)
+	brokerSecret, err := secret.EnsureFromSA(client, constants.SubmarinerBrokerAdminSA, "submariner-k8s-broker")
+	if err != nil {
+		status.Failure(err.Error())
+		return false
+	}
```
tpantelis: This function checks K8s version support so how is this code related? Also the "submariner-k8s-broker" namespace is hard-coded which is problematic.

tpantelis (dismissed):

pkg/diagnose/k8s_version.go
vthapar: > This function checks K8s version support so how is this code related? Also the "submariner-k8s-broker" namespace is hard-coded which is problematic.

This is not for fix, this was just an easy way to test the changes without having to fully deploy submariner while working on code changes.

vthapar:
@skitt  @tpantelis This is ready for review. I have only tweaked the UTs just enough to pass so that we can get this in and unblock https://github.com/submariner-io/shipyard/pull/825 I can add UTs in a subsequent PR, unless you think UTs need to come in this patch itself.

pkg/broker/ensure.go
```diff
@@ -204,6 +205,18 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
...
-	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Create(
-		context.TODO(), NewBrokerSA(submarinerBrokerSA), metav1.CreateOptions{})
+	sa := NewBrokerSA(submarinerBrokerSA)
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = serviceaccount.EnsureSecretFromSA(kubeClient, sa.Name, inNamespace)
```
skitt: I would prefer the secret handling to be hidden in the SA handling — callers shouldn’t have to know that to set up a service account correctly, they have to set up a secret.

skitt (changes requested):

pkg/serviceaccount/ensure.go
```diff
@@ -20,28 +20,154 @@ package serviceaccount
...
+
+	saSecret := getSecretFromSA(client, sa)
+
+	if saSecret != nil {
+		return saSecret, nil
+	}
+
+	// We couldn't find right secret from this SA, search all Secrets
+	saSecret, err = getSecretForSA(client, sa)
+	if err != nil {
```
tpantelis, line 128: We should distinguish the `NotFound` error.

```suggestion
	if err != nil && !apierrors.IsNotFound(err) {
	    return err
	}
	
	if err != nil {
	
```

tpantelis (changes requested):

nyechiel:
@vthapar is this ready for review? It looks like two CI jobs are failing

vthapar:
> @vthapar is this ready for review? It looks like two CI jobs are failing

Upgrade job is the only one that failed more than once, e2e has failed just once. Re-ran the jobs to see if e2e fails again. Looking into upgrade job issue.

dfarrell07 (approved):

Jaanki (approved):

nyechiel:
@skitt @sridhargaddam @tpantelis this one needs review

sridhargaddam (approved):

pkg/broker/ensure.go
```diff
@@ -204,6 +205,18 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
...
-	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Create(
-		context.TODO(), NewBrokerSA(submarinerBrokerSA), metav1.CreateOptions{})
+	sa := NewBrokerSA(submarinerBrokerSA)
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa, true)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = serviceaccount.EnsureSecretFromSA(kubeClient, sa.Name, inNamespace)
```
skitt: My previous comment still applies, unless I’ve missed something:

> I would prefer the secret handling to be hidden in the SA handling — callers shouldn’t have to know that to set up a service account correctly, they have to set up a secret.

In other words, there should only be a call to `serviceaccount.Ensure()` in this function, and `serviceaccount.Ensure()` should handle the secret (perhaps by calling `EnsureSecretFromSA()`, of course, but that should be transparent to the caller).

skitt (changes requested):

nyechiel:
@vthapar can you please address Stephen's comment?

pkg/broker/ensure.go
vthapar: > My previous comment still applies, unless I’ve missed something:
> 
> > I would prefer the secret handling to be hidden in the SA handling — callers shouldn’t have to know that to set up a service account correctly, they have to set up a secret.
> 
> In other words, there should only be a call to `serviceaccount.Ensure()` in this function, and `serviceaccount.Ensure()` should handle the secret (perhaps by calling `EnsureSecretFromSA()`, of course, but that should be transparent to the caller).

Got it. I missed the context of comment for `Ensure` and only addressed for `EnsureFromYaml`.

vthapar:
> @vthapar can you please address Stephen's comment?

Done.

pkg/serviceaccount/ensure.go
```diff
@@ -20,28 +20,182 @@ package serviceaccount
...
 }
 
 // nolint:wrapcheck // No need to wrap errors here.
-func Ensure(kubeClient kubernetes.Interface, namespace string, sa *corev1.ServiceAccount) (bool, error) {
+func ensure(kubeClient kubernetes.Interface, namespace string, sa *corev1.ServiceAccount, onlyCreate bool) (bool, error) {
+	if onlyCreate {
+		_, err := kubeClient.CoreV1().ServiceAccounts(namespace).Get(context.TODO(), sa.Name, metav1.GetOptions{})
+		if err == nil {
+			return true, nil
+		}
```
tpantelis: We should distinguish `NotFound` error

```suggestion
                 if apierrors.IsNotFound(err)
			return nil
		}
		
		if err != nil {
			return err
		}
```

pkg/serviceaccount/ensure.go
```diff
@@ -20,28 +20,182 @@ package serviceaccount
...
+		return nil, err
 	}
 
-	return Ensure(kubeClient, namespace, sa)
+	return sa, err
 }
 
 // nolint:wrapcheck // No need to wrap errors here.
-func Ensure(kubeClient kubernetes.Interface, namespace string, sa *corev1.ServiceAccount) (bool, error) {
+func ensure(kubeClient kubernetes.Interface, namespace string, sa *corev1.ServiceAccount, onlyCreate bool) (bool, error) {
```
tpantelis: Callers don't use the bool return value so we can remove it to simplify.

tpantelis (changes requested):

tpantelis (approved):

pkg/broker/ensure.go
```diff
@@ -203,7 +204,12 @@ func CreateNewBrokerRoleBinding(kubeClient kubernetes.Interface, serviceAccount,
...
-func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (brokerSA *v1.ServiceAccount, err error) {
-	return kubeClient.CoreV1().ServiceAccounts(inNamespace).Create(
-		context.TODO(), NewBrokerSA(submarinerBrokerSA), metav1.CreateOptions{})
+func CreateNewBrokerSA(kubeClient kubernetes.Interface, submarinerBrokerSA, inNamespace string) (err error) {
+	sa := NewBrokerSA(submarinerBrokerSA)
+	_, err = serviceaccount.Ensure(kubeClient, inNamespace, sa, true)
+
+	if apierrors.IsNotFound(err) {
+
+	}
```
skitt: There’s something missing here, I presume.

pkg/serviceaccount/ensure.go
```diff
@@ -20,28 +20,184 @@ package serviceaccount
...
+	// We couldn't find right secret from this SA, search all Secrets
+	saSecret, err = getSecretForSA(client, sa)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if err != nil {
+		newSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("%s-token-%s", sa.Name, generateRandomString(5)),
```
skitt, line 131: Can’t we use `GenerateName` here?

skitt (changes requested):

pkg/serviceaccount/ensure.go
skitt: We can fix this up later, this shouldn’t block the PR.

pkg/broker/ensure.go
vthapar: > There’s something missing here, I presume.

Missed cleanup after refactor.

skitt (approved):

pkg/serviceaccount/ensure.go
vthapar: I used GenerateName in intial draft, but then needed to do another Get to find the generated name. Went with this coz similar to what OCP's controller is doing.

PR #136, opened 2022-06-28
Gather creates a subdirectory with cluster name for gather logs

### LINKED ISSUE #54: subctl gather: Create sub-directory for each cluster

For `subctl gather`, create a sub-directory for each cluster instead of embedding the cluster name  in the file name.

**Why is this needed**:

The consensus is this will make it easier to review and search the files when troubleshooting.

skitt:
- this commit creates a subdirectory name derived from kubeconfig
  cluster.Name for gather logs to be stored in separate directories
  under the root timestamped directory

Resolves #54 

Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>

### DIFF
--- a/internal/gather/gather.go
+++ b/internal/gather/gather.go
@@ -23,7 +23,7 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"time"
+	"path/filepath"
 
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/admiral/pkg/stringset"
@@ -72,9 +72,8 @@ func Data(clusterInfo *cluster.Info, status reporter.Interface, options Options)
 		Deduplicate: true,
 	}))
 
-	if options.Directory == "" {
-		options.Directory = "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
-	}
+	// concatenate the name of the cluster with the root gather directory
+	options.Directory = filepath.Join(options.Directory, clusterInfo.Name)
 
 	if _, err := os.Stat(options.Directory); os.IsNotExist(err) {
 		err := os.MkdirAll(options.Directory, 0o700)
--- a/scripts/test/lib_subctl_gather_test.sh
+++ b/scripts/test/lib_subctl_gather_test.sh
@@ -61,7 +61,7 @@ function validate_pod_log_files() {
   read -ra pod_names_array <<< "$pod_names"
 
   for pod_name in "${pod_names_array[@]}"; do
-    file=$gather_out_dir/${cluster}_$pod_name.log
+    file=$gather_out_dir/${cluster}/$pod_name.log
     cat $file
 
   done
@@ -96,7 +96,7 @@ function validate_resource_files() {
   for i in "${!names_array[@]}"; do
     name=${names_array[$i]}
     namespace=${namespaces_array[$i]}
-    file=$gather_out_dir/${cluster_name}_${short_res}_${namespace}_${name}.yaml
+    file=$gather_out_dir/${cluster_name}/${short_res}_${namespace}_${name}.yaml
     cat $file
 
     kind_count=$(grep "kind: $kind$" $file | wc -l)
@@ -114,8 +114,8 @@ function validate_resource_files() {
 }
 
 function validate_broker_resources() {
-  validate_resource_files $submariner_broker_ns 'endpoints.submariner.io' 'Endpoint' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'clusters.submariner.io' 'Cluster' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'serviceimports.multicluster.x-k8s.io' 'ServiceImport' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'endpointslices.discovery.k8s.io' 'EndpointSlice' '' 'broker'
+  validate_resource_files $submariner_broker_ns endpoints.submariner.io Endpoint
+  validate_resource_files $submariner_broker_ns clusters.submariner.io Cluster
+  validate_resource_files $submariner_broker_ns serviceimports.multicluster.x-k8s.io ServiceImport
+  validate_resource_files $submariner_broker_ns endpointslices.discovery.k8s.io EndpointSlice
 }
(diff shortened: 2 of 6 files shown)


internal/gather/gather.go
```diff
@@ -76,6 +76,9 @@ func Data(clusterInfo *cluster.Info, status reporter.Interface, options Options)
 		options.Directory = "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
 	}
 
+	// concatenate the name of the cluster with the root gather directory
+	options.Directory = fmt.Sprintf("%s/%s", options.Directory, clusterInfo.Name)
```
skitt: This should use Go’s path construction functions instead of concatenating strings directly.

Jaanki:
system test is failing because https://github.com/submariner-io/subctl/blob/393ddf864a1113d95f4dd242e475ab82e341fc26/scripts/test/system.sh#L46 needs to be changed too to reflect the actual cluster name "cluster1" and not "broker".

mkolesnik (changes requested):
Since the directory names are now cluster names, please also change https://github.com/submariner-io/subctl/blob/devel/internal/gather/resource.go#L60 to not contain cluster name.

scripts/test/lib_subctl_gather_test.sh
```diff
@@ -61,7 +61,7 @@ function validate_pod_log_files() {
   read -ra pod_names_array <<< "$pod_names"
 
   for pod_name in "${pod_names_array[@]}"; do
-    file=$gather_out_dir/${cluster}_$pod_name.log
+    file=$gather_out_dir/${cluster}/${cluster}_$pod_name.log
```
mkolesnik:
```suggestion
    file="${gather_out_dir}/${cluster}/${pod_name}.log"
```

Let's instead omit cluster name from the log file name in https://github.com/submariner-io/subctl/blob/devel/internal/gather/logs.go#L105

scripts/test/system.sh
```diff
@@ -42,8 +42,8 @@ function test_subctl_gather() {
         with_context "${cluster}" validate_gathered_files
     done
 
-    # Broker
-    with_context "$broker" validate_broker_resources
+    # Broker (on the first cluster)
+    with_context "${clusters[0]}" validate_broker_resources
```
mkolesnik: No need to change this, `$broker` is determined by Shipyard's `load_settings` and thus we can use it here.

Unless I'm missing something?

mkolesnik: Re the system tests you'll need to update how it gathers broker info and tests it, as clearly there was some special expectation re the broker, eg https://github.com/submariner-io/subctl/blob/devel/internal/gather/gather.go#L225-L231

internal/gather/gather.go
```diff
@@ -76,6 +77,9 @@ func Data(clusterInfo *cluster.Info, status reporter.Interface, options Options)
 		options.Directory = "submariner-" + time.Now().UTC().Format("20060102150405") // submariner-YYYYMMDDHHMMSS
 	}
 
+	// concatenate the name of the cluster with the root gather directory
+	options.Directory = filepath.Join(options.Directory, clusterInfo.Name)
```
mkolesnik, line 78: Perhaps this should be skipped in case you're only gathering from a single cluster?

Otherwise the first directory will just have another nested in it, making it a bit useless.

skitt: I think it’s worth keeping the cluster directory even if it’s the only directory, since the cluster name no longer appears in the generated file names. It also means the output is consistent regardless of how many clusters are “gathered”, and multiple “gather” directories for different clusters can easily be merged.

nyechiel:
@skitt I believe this is the [parent issue](https://github.com/submariner-io/subctl/issues/54)? I think we better track the issue in the board and not the PR.

scripts/test/system.sh
```diff
@@ -42,8 +42,8 @@ function test_subctl_gather() {
         with_context "${cluster}" validate_gathered_files
     done
 
-    # Broker
-    with_context "$broker" validate_broker_resources
+    # Broker (on the first cluster)
+    with_context "${clusters[0]}" validate_broker_resources "${clusters[0]}"
```
mkolesnik:
```suggestion
    with_context "${broker}" validate_broker_resources
```

`$broker` is set when loading the settings, no reason not to use it

scripts/test/lib_subctl_gather_test.sh
```diff
@@ -114,8 +114,8 @@ function validate_resource_files() {
...
 
 function validate_broker_resources() {
-  validate_resource_files $submariner_broker_ns 'endpoints.submariner.io' 'Endpoint' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'clusters.submariner.io' 'Cluster' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'serviceimports.multicluster.x-k8s.io' 'ServiceImport' '' 'broker'
-  validate_resource_files $submariner_broker_ns 'endpointslices.discovery.k8s.io' 'EndpointSlice' '' 'broker'
+  validate_resource_files $submariner_broker_ns 'endpoints.submariner.io' 'Endpoint' '' "$1"
+  validate_resource_files $submariner_broker_ns 'clusters.submariner.io' 'Cluster' '' "$1"
+  validate_resource_files $submariner_broker_ns 'serviceimports.multicluster.x-k8s.io' 'ServiceImport' '' "$1"
+  validate_resource_files $submariner_broker_ns 'endpointslices.discovery.k8s.io' 'EndpointSlice' '' "$1"
```
mkolesnik:
```suggestion
  validate_resource_files $submariner_broker_ns 'endpoints.submariner.io' 'Endpoint'
  validate_resource_files $submariner_broker_ns 'clusters.submariner.io' 'Cluster'
  validate_resource_files $submariner_broker_ns 'serviceimports.multicluster.x-k8s.io' 'ServiceImport'
  validate_resource_files $submariner_broker_ns 'endpointslices.discovery.k8s.io' 'EndpointSlice'
```

4th arg is "selector" which can be omitted, 5th arg defaults to `$cluster` which is set globally by `with_context`

mkolesnik (changes requested):

scripts/test/lib_subctl_gather_test.sh
mkolesnik: @skitt kind reminder :point_up:

skitt: Ah yes, thanks!

mkolesnik (approved):

tpantelis (approved):

PR #152, opened 2022-07-07
Remove K8s E2E param, use default

dfarrell07:
Instead of explicitly setting the K8s version for E2E-Full CI, just use
the default from Shipyard. We typically want to use the same latest-K8s
version in all repos once we test it and make it available in Shipyard.
This will require less maintenance.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -15,8 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
-        # Run tests against the latest K8s version
-        k8s_version: ['1.23']
         lighthouse: ['', 'lighthouse']
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-full.yml
```diff
@@ -16,7 +16,6 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         # Run tests against the latest K8s version
```
mkolesnik:
```suggestion
```

mkolesnik (changes requested):

mkolesnik (approved):

skitt (approved):

dfarrell07:
The GitHub required jobs config needs to be updated by an admin.

PR #156, opened 2022-07-11
Rework join kubecontext handling using clientcmd

skitt:
This enables all available connection options. We can also use
clientcmd's built-in support for disabling TLS certificate checks,
instead of our own --check-broker-certificate.

### DIFF
--- a/cmd/subctl/aws.go
+++ b/cmd/subctl/aws.go
@@ -23,6 +23,7 @@ import (
 	cpaws "github.com/submariner-io/cloud-prepare/pkg/aws"
 	"github.com/submariner-io/subctl/internal/cli"
 	"github.com/submariner-io/subctl/internal/exit"
+	"github.com/submariner-io/subctl/internal/restconfig"
 	cloudaws "github.com/submariner-io/subctl/pkg/cloud/aws"
 	"github.com/submariner-io/subctl/pkg/cloud/cleanup"
 	"github.com/submariner-io/subctl/pkg/cloud/prepare"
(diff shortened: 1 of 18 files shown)


skitt:
The resulting flags are

```
Connect a cluster to an existing broker

Usage:
  subctl join [flags]

Flags:
      --as string                                 Username to impersonate for the operation
      --as-group stringArray                      Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --as-uid string                             UID to impersonate for the operation
      --cable-driver string                       cable driver implementation
      --certificate-authority string              Path to a cert file for the certificate authority
      --client-certificate string                 Path to a client certificate file for TLS
      --client-key string                         Path to a client key file for TLS
      --cluster string                            The name of the kubeconfig cluster to use
      --clustercidr string                        cluster CIDR
      --clusterid string                          cluster ID used to identify the tunnels
      --context string                            The name of the kubeconfig context to use
      --coredns-custom-configmap string           Name of the custom CoreDNS configmap to configure forwarding to lighthouse. It should be in <namespace>/<name> format where <namespace> is optional and defaults to kube-system
      --custom-domains strings                    list of domains to use for multicluster service discovery
      --force-udp-encaps                          force UDP encapsulation for IPSec
      --globalnet                                 enable/disable Globalnet for this cluster (default true)
      --globalnet-cidr string                     GlobalCIDR to be allocated to the cluster
      --globalnet-cluster-size uint               cluster size for GlobalCIDR allocated to this cluster (amount of global IPs)
      --health-check                              enable Gateway health check (default true)
      --health-check-interval uint                interval in seconds between health check packets (default 1)
      --health-check-max-packet-loss-count uint   maximum number of packets lost before the connection is marked as down (default 5)
  -h, --help                                      help for join
      --ignore-requirements                       ignore requirement failures (unsupported)
      --image-override strings                    override component image
      --insecure-skip-tls-verify                  If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --ipsec-debug                               enable IPsec debugging (verbose logging)
      --kubeconfig string                         absolute path(s) to the kubeconfig file(s)
      --label-gateway                             label gateways if necessary (default true)
      --load-balancer                             enable automatic LoadBalancer in front of the gateways
  -n, --namespace string                          If present, the namespace scope for this CLI request
      --natt                                      enable NAT traversal for IPsec (default true)
      --nattport int                              IPsec NATT port (default 4500)
      --operator-debug                            enable operator debugging (verbose logging)
      --password string                           Password for basic authentication to the API server
      --pod-debug                                 enable Submariner pod debugging (verbose logging in the deployed pods)
      --preferred-server                          enable this cluster as a preferred server for dataplane connections
      --repository string                         image repository
      --request-timeout string                    The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
      --server string                             The address and port of the Kubernetes API server
      --servicecidr string                        service CIDR
      --tls-server-name string                    If provided, this name will be used to validate server certificate. If this is not provided, hostname used to contact the server is used.
      --token string                              Bearer token for authentication to the API server
      --user string                               The name of the kubeconfig user to use
      --username string                           Username for basic authentication to the API server
      --version string                            image version
```

PR #171, opened 2022-07-18
Mention default cable driver in join help msg

### LINKED ISSUE #166: `subctl join --help` should display default cable driver
**What happened**:
help on `subctl join` should mention that the default "cable driver" is `libreswan`.

```
$ subctl join --help
Connect a cluster to an existing broker

Usage:
  subctl join [flags]

Flags:
      --cable-driver string                       cable driver implementation
```

The message should be somthing like this:

```
Flags:
      --cable-driver string                       cable driver implementation (default libreswan)
```

**Environment**:
- Install tools:
- Others:

Jaanki:
While joining a cluster, if the cable driver is not mentioned, the
default is `libreswan`. This information should be displayed while
running `subctl join --help` like this:

```
$ cmd/bin/subctl-devel-c9c8c8c-linux-amd64 join --help
Connect a cluster to an existing broker

Usage:
  subctl join [flags]

Flags:
      --cable-driver string                       cable driver
implementation (default "libreswan")
```

Closes: #166

### DIFF
--- a/cmd/subctl/join.go
+++ b/cmd/subctl/join.go
@@ -108,7 +108,7 @@ func addJoinFlags(cmd *cobra.Command) {
 		"enable Submariner pod debugging (verbose logging in the deployed pods)")
 	cmd.Flags().BoolVar(&joinFlags.OperatorDebug, "operator-debug", false, "enable operator debugging (verbose logging)")
 	cmd.Flags().BoolVar(&labelGateway, "label-gateway", true, "label gateways if necessary")
-	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "", "cable driver implementation")
+	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "libreswan", "cable driver implementation")
 	cmd.Flags().UintVar(&joinFlags.GlobalnetClusterSize, "globalnet-cluster-size", 0,
 		"cluster size for GlobalCIDR allocated to this cluster (amount of global IPs)")
 	cmd.Flags().StringVar(&joinFlags.GlobalnetCIDR, "globalnet-cidr", "",


tpantelis (approved):

maayanf24 (dismissed):

skitt (changes requested):
Currently the default isn’t set by `subctl` or even the operator, it’s determined by the gateway (https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L50).

If we change this, the default will be set by `subctl` for `subctl` deployments; that’s a reasonable change, but we need to ensure that the ramifications have been checked. In particular, we need to ensure that tests which rely on other cable drivers still work with this. (I don’t think there’s a risk, but we need to check.)

Jaanki:
> Currently the default isn’t set by `subctl` or even the operator, it’s determined by the gateway (https://github.com/submariner-io/submariner/blob/devel/pkg/cable/libreswan/libreswan.go#L50).
> 
> If we change this, the default will be set by `subctl` for `subctl` deployments; that’s a reasonable change, but we need to ensure that the ramifications have been checked. In particular, we need to ensure that tests which rely on other cable drivers still work with this. (I don’t think there’s a risk, but we need to check.)

We should add the full test matrix GHA we have on operator here as well. That would help in testing such PRs in future too. Opened an issue to track this https://github.com/submariner-io/subctl/issues/175

tpantelis:
We can compromise by appending _(default "libreswan")_ to our help text instead of actually setting it as the default value. To the end user the effect is the same.

cmd/subctl/join.go
```diff
@@ -113,7 +113,7 @@ func addJoinFlags(cmd *cobra.Command) {
 		"enable Submariner pod debugging (verbose logging in the deployed pods)")
 	cmd.Flags().BoolVar(&joinFlags.OperatorDebug, "operator-debug", false, "enable operator debugging (verbose logging)")
 	cmd.Flags().BoolVar(&labelGateway, "label-gateway", true, "label gateways if necessary")
-	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "", "cable driver implementation")
+	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "libreswan", "cable driver implementation")
```
tpantelis, line 111: We can compromise by mentioning the default in our help text instead of actually setting it as the default value. To the end user the effect is the same.

```suggestion
	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "", "cable driver implementation (default "libreswan")")
```

tpantelis (changes requested):

cmd/subctl/join.go
skitt: That works for me! No behaviour change in the code, and the user still has the important information.

Jaanki: Ya. Thats a good idea.

sridhargaddam (dismissed):

cmd/subctl/join.go
```diff
@@ -113,7 +113,7 @@ func addJoinFlags(cmd *cobra.Command) {
 		"enable Submariner pod debugging (verbose logging in the deployed pods)")
 	cmd.Flags().BoolVar(&joinFlags.OperatorDebug, "operator-debug", false, "enable operator debugging (verbose logging)")
 	cmd.Flags().BoolVar(&labelGateway, "label-gateway", true, "label gateways if necessary")
-	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "", "cable driver implementation")
+	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "", "cable driver implementation (default \"libreswan\")")
```
sridhargaddam: There is a small typo in the commit message. If you happen to respin, please fix it.

drvier -> driver

mkolesnik:
```suggestion
	cmd.Flags().StringVar(&joinFlags.CableDriver, "cable-driver", "libreswan", "cable driver implementation")
```

I agree with @skitt here, if you want a default value set it explicitly here (and as a bonus it also gets printed).

mkolesnik (dismissed):

skitt (dismissed):

tpantelis (dismissed):

cmd/subctl/join.go
skitt: The initial version of this PR set the default, and I urged caution since that would involve changing the default codepaths — I didn’t think there was much risk but it would have meant ensuring that all our tests were still OK. Adding the default in the help text allows us to provide the same information without changing the default codepaths.

mkolesnik: I don't see how that makes any sense actually, if we leave it empty then the default is determined elsewhere and could be different than the "documented" one and as such misleading.

If anything, the safer choice is to set the default here regardless of what the default is in the underlying implementation.

As for tests, theres no real danger in one way or the other, since a default "libreswan" is set somewhere, and if we fear of that then we should set it explicitly, but TBH it doesnt matter whats the default in that case as its all expected to still work.

mkolesnik (commented):
Lets finish the discussion

tpantelis (approved):

skitt (approved):

PR #172, opened 2022-07-18
Allow passing DEFAULT_REPO while building subctl

Jaanki:
This makes the repo to pull images from configurable while building
subctl.

This is useful for developers who want to test their images from a
registry other than quay.io/submariner. Our downstream builds is one
such example where the users need a different default repo to pull the
images from.

Usage:

make cmd/bin/subctl DEFAULT_REPO=<quay.io/devaccount>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ DEFAULT_IMAGE_VERSION ?= $(BASE_BRANCH)
 export BASE_BRANCH
 export DEFAULT_IMAGE_VERSION
 
+DEFAULT_REPO ?= "quay.io/submariner"
+
 # Define LOCAL_BUILD to build directly on the host and not inside a Dapper container
 ifdef LOCAL_BUILD
 DAPPER_HOST_ARCH ?= $(shell go env GOHOSTARCH)
@@ -117,7 +119,8 @@ cmd/bin/subctl-%: $(shell find . -name "*.go") $(VENDOR_MODULES)
 	export GOARCH GOOS; \
 	$(SCRIPTS_DIR)/compile.sh \
 		--ldflags "-X 'github.com/submariner-io/subctl/pkg/version.Version=$(VERSION)' \
-		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}'" \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}' \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultRepo=$(DEFAULT_REPO)'" \
         --noupx $@ ./cmd $(BUILD_ARGS)
 
 ci: golangci-lint markdownlint unit build images


tpantelis:
What is the use case for this, ie how and where would this be used?

Makefile
```diff
@@ -4,6 +4,9 @@ DEFAULT_IMAGE_VERSION ?= $(BASE_BRANCH)
 export BASE_BRANCH
 export DEFAULT_IMAGE_VERSION
 
+DEFAULT_REPO = "quay.io/submariner"
+export DEFAULT_REPO
```
skitt: This isn’t necessary since the variable is only used in the Makefile:

```suggestion
```

skitt (changes requested):
I have the same question as Tom. Please rephrase the commit message to describe the feature you’re adding rather than how you’re going about it.

Makefile
skitt: That’s wrong, we’re using the shell variable (`$${…}`) so we do need to export it.

Makefile
```diff
@@ -117,7 +120,8 @@ cmd/bin/subctl-%: $(shell find . -name "*.go") $(VENDOR_MODULES)
 	export GOARCH GOOS; \
 	$(SCRIPTS_DIR)/compile.sh \
 		--ldflags "-X 'github.com/submariner-io/subctl/pkg/version.Version=$(VERSION)' \
-		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}'" \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}' \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultRepo=$${DEFAULT_REPO#v}'" \
```
skitt: If you change this to

```suggestion
		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultRepo=$(DEFAULT_REPO)'" \
```

then the `export` can be dropped. We need to export `DEFAULT_IMAGE_VERSION` because we rely on a shell feature to post-process it (`#v`), but we don’t need that for `DEFAULT_REPO` so we can use Make variables only.

skitt (changes requested):

Jaanki:
Updated the PR description and commit message with the use case.

skitt:
> Updated the PR description and commit message with the use case.

Thanks, but the commit message is still describing the implementation rather than the feature. I was thinking of something more like

> Allow the default image repo to be configured during build
>
> This allows the default image repo to be specified by setting the DEFAULT_REPO variable.
>
> This is useful for developers who want to test their images from a hub
other than quay.io/submariner:
>
>     make ... DEFAULT_REPO=quay.io/devaccount

Makefile
```diff
@@ -117,7 +119,8 @@ cmd/bin/subctl-%: $(shell find . -name "*.go") $(VENDOR_MODULES)
 	export GOARCH GOOS; \
 	$(SCRIPTS_DIR)/compile.sh \
 		--ldflags "-X 'github.com/submariner-io/subctl/pkg/version.Version=$(VERSION)' \
-		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}'" \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=$${DEFAULT_IMAGE_VERSION#v}' \
+		       -X 'github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultRepo=${DEFAULT_REPO}'" \
```
skitt: `DEFAULT_REPO` needs to be surrounded by parentheses, not curly braces, see https://github.com/submariner-io/subctl/pull/172#discussion_r924263455

skitt (changes requested):

mkolesnik (changes requested):
I don't understand why this is needed at all, we already have flags [1] to specify a custom repo while joining and/or specific image overrides, why introduce another way to do exactly the same thing?

[1] https://submariner.io/operations/deployment/subctl/#join-flags-images-and-repositories

Jaanki:
> I don't understand why this is needed at all, we already have flags [1] to specify a custom repo while joining and/or specific image overrides, why introduce another way to do exactly the same thing?
> 
> [1] https://submariner.io/operations/deployment/subctl/#join-flags-images-and-repositories

This allows a `subctl` binary to be built referencing a specific repository other than the default, whether for test purposes or for distribution purposes. That binary can then be provided to users without having to remind them to specify the different options for the repository.

mkolesnik:
> This allows a `subctl` binary to be built referencing a specific repository other than the default, whether for test purposes or for distribution purposes. That binary can then be provided to users without having to remind them to specify the different options for the repository.

That doesn't seem useful as of now, given that flags can easily be used to specify any repo.

skitt:
> > This allows a `subctl` binary to be built referencing a specific repository other than the default, whether for test purposes or for distribution purposes. That binary can then be provided to users without having to remind them to specify the different options for the repository.
> 
> That doesn't seem useful as of now, given that flags can easily be used to specify any repo.

The important feature is this part: “without having to remind them to specify the different options for the repository”. If you want a specific use case, this would be used for the downstream build — to build a `subctl` binary which references the downstream repositories by default instead of the upstream repositories.

mkolesnik:
> The important feature is this part: “without having to remind them to specify the different options for the repository”. If you want a specific use case, this would be used for the downstream build — to build a `subctl` binary which references the downstream repositories by default instead of the upstream repositories.

Downstream (or generally forked) builds are a much better justification than "This is useful for developers who want to test their images", as the "testing" part can be easily achieved with the flags in the current implementation.

The commit message should reflect this IMHO as it's a case where you need the default repo different even for users of the binary.

Makefile
```diff
@@ -4,6 +4,8 @@ DEFAULT_IMAGE_VERSION ?= $(BASE_BRANCH)
 export BASE_BRANCH
 export DEFAULT_IMAGE_VERSION
 
+DEFAULT_REPO = "quay.io/submariner"
```
mkolesnik:
```suggestion
DEFAULT_REPO ?= "quay.io/submariner"
```

Won't this be better, to allow builder to dynamically specify the repo during build time?

Makefile
mkolesnik: Technically both should work according to the docs [1] but since we want to avoid confusion by devs it's probably best to use parenthesis to explicitly signal it's a make variable.

[1] https://www.gnu.org/software/make/manual/html_node/Reference.html#Reference

Makefile
skitt: `make DEFAULT_REPO=…` overrides this as-is; `?=` is useful to keep an existing variable value that might have been set in a Makefile (_e.g._ through an `include`). Since this is at the start of the main `subctl` Makefile, it’s not all that useful (but it wouldn’t hurt).

mkolesnik: AFAIK `?=` supports values from the env while `=` would overwrite them, so I still think it'll be more useful.

e.g. `DEFAULT_REPO=… make ...` won't work unless it's assigned via `?=`

skitt: Ah yes, that’s a good point, so `?=` would indeed be better.

Jaanki: well, `make DEFAULT_REPO="quay.io/submariner2" cmd/bin/subctl` works fine 
`Building 'cmd/bin/subctl-devel-db0dc08-linux-amd64' (ldflags: $'-X github.com/submariner-io/subctl/pkg/version.Version=devel-db0dc08 \t       -X github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultSubmarinerOperatorVersion=devel \t       -X github.com/submariner-io/submariner-operator/api/submariner/v1alpha1.DefaultRepo=quay.io/submariner2')`

without the `?=` but I have anyways assigned it like that.

mkolesnik: > without the `?=` but I have anyways assigned it like that.

Where? The current PR doesnt reflect that

Jaanki: Now it does.

mkolesnik (approved):

skitt (approved):

tpantelis (approved):

PR #173, opened 2022-07-18
Display proper output message on `subctl show brokers`

### LINKED ISSUE #167: `subctl show brokers` output empty on non-broker cluster
**What happened**:
On a non-broker cluster (cluster2 in this case), `subctl show brokers` doesn't print anything. Just a blank line.
```
$ subctl show brokers
Cluster "cluster1"
 ✓ Detecting broker(s)
NAMESPACE                NAME                     COMPONENTS                        GLOBALNET ENABLED   GLOBALNET CIDR       DEFAULT GLOBALNET SIZE    DEFAULT DOMAINS                         
submariner-k8s-broker    submariner-broker        service-discovery, connectivity   false               242.0.0.0/8          65536                                                             

Cluster "cluster2"
 ✓ Detecting broker(s)


```

It should print something meaningful like "Broker is not installed on this cluster".

Run `subctl show brokers`

This also affects `subctl show all`

**Environment**:
- Install tools:
- Others:

Jaanki:
when broker is not installed on the cluster.

Currently a blank line is displayed. Instead display that "No brokers
found" when a broker is not installed on the cluster.

The output will look like

$ cmd/bin/subctl-devel-59cc4f5-linux-amd64 show brokers
Cluster "cluster2"
 ✓ Detecting broker(s)
 ✓ No brokers found

Closes: #167

### DIFF
--- a/internal/show/brokers.go
+++ b/internal/show/brokers.go
@@ -45,6 +45,7 @@ func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
 
 	brokers := brokerList.Items
 	if len(brokers) == 0 {
+		status.Success("No brokers found")
 		return true
 	}
 


internal/show/brokers.go
```diff
@@ -45,6 +46,7 @@ func Brokers(clusterInfo *cluster.Info, status reporter.Interface) bool {
 
 	brokers := brokerList.Items
 	if len(brokers) == 0 {
+		fmt.Println("   Broker not installed")
```
dfarrell07: I see your sample output in the commit message, and to me it looks like it would need an extra space to align. I wonder if we need the complexity of indentation at all? :shrug:

tpantelis: Use the status to report the message.

```suggestion
		status.Success("No brokers found")
```

tpantelis (changes requested):

internal/show/brokers.go
Jaanki: While trying locally, it showed up correctly. Indentation looks nice. But then all other outputs (except for show network) don't have indentation. I am also wondering if we can use status.success here.

Jaanki: "No brokers found" was my initial thought. But not installed sounds more accurate.

Jaanki: Or should we use `status.Warning` like it is done when `submariner` is not installed in the cluster?

Jaanki: @tpantelis @skitt WDYT ^ ?

tpantelis: A warning indicates it could be a problem but it's not required for a broker to be installed so this doesn't qualify as a warning.

sridhargaddam (approved):

mkolesnik (approved):

tpantelis (approved):

PR #190, opened 2022-07-28
Use GroupVersion.WithResource

skitt:
This simplifies GroupVersionResource construction code.


tpantelis (approved):

internal/gather/servicediscovery.go
```diff
@@ -49,19 +48,11 @@ func gatherCoreDNSPodLogs(info *Info) {
 }
 
 func gatherServiceExports(info *Info, namespace string) {
-	ResourcesToYAMLFile(info, schema.GroupVersionResource{
-		Group:    mcsv1a1.GroupName,
-		Version:  mcsv1a1.GroupVersion.Version,
-		Resource: "serviceexports",
-	}, namespace, metav1.ListOptions{})
+	ResourcesToYAMLFile(info, mcsv1a1.SchemeGroupVersion.WithResource("serviceexports"), namespace, metav1.ListOptions{})
```
tpantelis: `mcsv1a1.SchemeGroupVersion` is actually deprecated - I'm guessing newer versions of the code generator don't generate it but the authors kept it for backwards compatibility. I think we should leave this as is - also in other places.

tpantelis (changes requested):

internal/gather/servicediscovery.go
skitt: They still do, the comment was added following https://github.com/kubernetes/kubernetes/pull/66251#discussion_r203715354 but I don’t know why `SchemeGroupVersion` is supposed to be deprecated (which puts a damper on this whole PR really).

tpantelis: yeah and someone asked that over a year ago but wasn't answered. We still need to use `SchemeGroupVersion` for the ones where the code has not been re-generated using a newer generator that deprecates it. So I think this PR is still valid, ie we might as well use `WithResource` where we can w/o being warned about deprecation.

internal/gvr/resource.go
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+package gvr
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func ResourceFromMetaGroupVersion(gv metav1.GroupVersion, resource string) schema.GroupVersionResource {
```
tpantelis: I think we can shorten this.

That said, I'm planning to change `gather` to use a controller client instead of a dynamic client in which case we would use GVK instead of GVR. Perhaps we can hold off on these changes for now.

```suggestion
func FromGroupVersion(gv metav1.GroupVersion, resource string) schema.GroupVersionResource {
```

tpantelis (changes requested):

internal/gvr/resource.go
tpantelis: I've decided to keep the dynamic client for ease of use so this change is relevant.

skitt: I’ve switched to `FromMetaGroupVersion` instead of `ResourceFrom…` because we’re specifically handling `meta` `GroupVersion` as opposed to `schema`; I’d rather keep the `Meta` reference in case we ever need to do something with one of the other versions.

tpantelis (approved):

internal/gvr/resource.go
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+package gvr
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func FromMetaGroupVersion(gv metav1.GroupVersion, resource string) schema.GroupVersionResource {
```
tpantelis, line 26: It would probably be useful to put this in **admiral** but we can do that later if needed.

dfarrell07 (approved):

PR #192, opened 2022-08-01
Convert uses of operator clientset to controller Client

tpantelis:
...except for `network.Discover` which is awaiting changes in **submariner-operator**.


skitt (approved):

skitt:
There’s a namespace missing somewhere but I haven’t figured out where from reviewing the change.

tpantelis:
> There’s a namespace missing somewhere but I haven’t figured out where from reviewing the change.

yeah 👀

dfarrell07 (approved):

PR #193, opened 2022-08-02
Remove use of client.Producer in internal/gather

tpantelis:
...in lieu of controller Client. However we still need the kube client set for subresource operations like get pod logs which aren't supported by the controller Client.


tpantelis:
I decided to go in a different direction and define a new `client.Producer`  and keep the dynamic client.

PR #220, opened 2022-08-18
Consolidate repo handling logic

mkolesnik:
This logic was spread around, consolidate it all into
`image.RepositoryInfo` so that it can be easily and consistently used.

Existing functionality isn't affected, as this is an internal refactor.


pkg/join/join.go
```diff
@@ -129,6 +133,33 @@ func ClusterToBroker(brokerInfo *broker.Info, options *Options, clientProducer c
...
+		}
+
+		imageOverrides[imageName] = override[1]
+	}
+
+	return imageOverrides, nil
+}
+
+func invalidImageName(key string) bool {
+	for _, name := range names.ValidImageNames {
```
tpantelis: As per [this discussion](https://github.com/submariner-io/submariner-operator/pull/2185#issuecomment-1219557264), we should validate against the component names. We can define a local `validComponentNames` var here.

tpantelis (approved):

skitt (approved):

PR #234, opened 2022-08-24
Add execute.IfServiceDiscoveryInstalled function

tpantelis:


cmd/subctl/execute/execute.go
```diff
@@ -75,3 +75,15 @@ func IfSubmarinerInstalled(run OnClusterFn) OnClusterFn {
 		return run(clusterInfo, status)
 	}
 }
+
+func IfServiceDiscoveryInstalled(run OnClusterFn) OnClusterFn {
+	return func(clusterInfo *cluster.Info, status reporter.Interface) bool {
+		if clusterInfo.ServiceDiscovery == nil {
+			status.Warning(constants.ServiceDiscoveryNotInstalled)
```
skitt, line 82: Do we always want to warn that service discovery is not installed? In some scenarios this would be perfectly fine, but some users may consider the warning to be an actual warning and indicative of a problem with their setup.

skitt (changes requested):

cmd/subctl/execute/execute.go
tpantelis: This is intended for top-level commands related to service discovery (eg diagnose) so I think warning makes sense over success or failure. This is what we do for Submariner not installed. We don't have an info message type.

skitt (approved):

LiavSlama (approved):

dfarrell07 (approved):

PR #242, opened 2022-08-30
clientcmd-based kubeconfig handling for single-context functions

skitt:
See individual commits for details.

This relates to https://github.com/submariner-io/enhancements/issues/79


internal/restconfig/restconfig.go
```diff
@@ -26,58 +26,134 @@ import (
...
+		_ = flags.MarkHidden("namespace")
+	}
+
+	rcp.defaultClientConfig = &configAndOverrides{
+		config:    clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &overrides),
+		overrides: &overrides,
+	}
+}
+
+type PerContextFn func(clusterName, namespace string, clientProducer client.Producer) error
```
skitt: I’ll rework this to use `cluster.Info`.

tpantelis: I assume in another PR?

skitt: No, in this PR.

skitt:
This works locally, I’m looking into why it fails in CI.

pkg/diagnose/connections.go
```diff
@@ -20,7 +20,7 @@ package diagnose
 
 import (
 	"github.com/submariner-io/admiral/pkg/reporter"
-	"github.com/submariner-io/subctl/pkg/cluster"
+	"github.com/submariner-io/subctl/internal/cluster"
```
tpantelis: `cluster.Info` is referenced in the public `Connections` function (and other public functions) so if it's moved internal I think it would be problematic for users outside the repo. Perhaps we can move the _internal/cluster_  functions to _pkg/cluster_ instead.

internal/restconfig/restconfig.go
```diff
@@ -26,58 +26,140 @@ import (
...
+		overrides: &overrides,
+	}
+}
+
+type PerContextFn func(clusterInfo *cluster.Info, namespace string) error
+
+// Runs the given function on the selected context.
+// The status reporter is only used for errors specific to this function;
+// the per-context function is expected to perform its own status reporting.
+func (rcp *Producer) RunOnSelectedContext(function PerContextFn, status reporter.Interface) error {
```
skitt, line 134: Since this ends up tied to a status reporter anyway, I wonder if I shouldn’t just add `status` to `PerContextFn`.

tpantelis: That seems reasonable.

pkg/cluster/cluster_test.go
```diff
@@ -23,7 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
```
tpantelis: Maybe a nit but since these functions as related to cluster IDs perhaps we could name the files accordingly, ie _id.go_ and _id_test.go_.

tpantelis (approved):

mkolesnik (commented):
@skitt do you reckon this PR needs the release note, or do you plan to add it on a later PR?

internal/restconfig/restconfig.go
```diff
@@ -52,24 +54,112 @@ type RestConfig struct {
...
+		if err != nil {
+			return status.Error(err, "error building the cluster.Info for the in-cluster configuration")
+		}
+
+		// In-cluster configurations don't specify a namespace, use the default
+		// When using the in-cluster configuration, that's the only configuration we want
+		return function(clusterInfo, "default", status)
+	}
+
+	if rcp.defaultClientConfig != nil {
```
mkolesnik: Shouldn't it error out if the config is not properly initialized?

I feel like we might miss a bug otherwise, as it'll just return no error but also not do anything AFAICT

internal/restconfig/restconfig.go
```diff
@@ -66,6 +66,7 @@ type Producer struct {
 	defaultClientConfig *configAndOverrides
 	inCluster           bool
 	namespaceFlag       bool
+	defaultNamespace    *string
```
mkolesnik, line 69: nit: Saving as pointer here is a kind of unnecessary optimization IMHO

cmd/subctl/deploybroker.go
```diff
@@ -40,31 +42,29 @@ var (
...
+		exit.OnError(deployRestConfigProducer.RunOnSelectedContext(
+			func(clusterInfo *cluster.Info, namespace string, status reporter.Interface) error {
+				if err := deploy.Broker(&deployflags, clusterInfo.ClientProducer, status); err != nil {
+					return err // nolint:wrapcheck // No need to wrap errors here.
+				}
+
+				return broker.WriteInfoToFile( // nolint:wrapcheck // No need to wrap errors here.
+					clusterInfo.RestConfig, deployflags.BrokerNamespace, ipsecSubmFile,
+					stringset.New(deployflags.BrokerSpec.Components...), deployflags.BrokerSpec.DefaultCustomDomains, status)
+			}, cli.NewReporter()))
```
mkolesnik: I find it relatively hard to read this anonymous function with such a high level of indentation.
Do you think we can extract it to a named function, for easier code maintainability?

cmd/subctl/join.go
```diff
@@ -55,34 +58,31 @@ var joinCmd = &cobra.Command{
...
-			joinFlags.CustomDomains = *brokerInfo.CustomDomains
-		}
+				if joinFlags.CustomDomains == nil && brokerInfo.CustomDomains != nil {
+					joinFlags.CustomDomains = *brokerInfo.CustomDomains
+				}
 
-		err = join.ClusterToBroker(brokerInfo, &joinFlags, clientProducer, status)
-		exit.OnError(err)
+				return join.ClusterToBroker( // nolint:wrapcheck // No need to wrap errors here.
+					brokerInfo, &joinFlags, clusterInfo.ClientProducer, status)
```
mkolesnik: I find it relatively hard to read this anonymous function with such a high level of indentation.
Do you think we can extract it to a named function, for easier code maintainability?

cmd/subctl/uninstall.go
```diff
@@ -23,55 +23,52 @@ import (
...
+
+					_ = survey.AskOne(prompt, &result)
+
+					if !result {
+						return nil
+					}
+				}
+
+				return uninstall.All( // nolint:wrapcheck // No need to wrap errors here.
+					clusterInfo.ClientProducer, clusterInfo.Name, namespace, status)
```
mkolesnik: I find it relatively hard to read this anonymous function with such a high level of indentation.
Do you think we can extract it to a named function, for easier code maintainability?

mkolesnik (commented):
Thanks for this :)

I had a few comments/questions

internal/restconfig/restconfig.go
skitt: Ah yes, I meant to address that earlier and forgot, thanks for noticing!

internal/restconfig/restconfig.go
skitt: It’s not an optimisation, it’s so that we can determine whether a value was given or not (_i.e._ `nil` is significant).

mkolesnik: Ah OK I found it, makes sense

cmd/subctl/unexport.go
```diff
@@ -41,35 +43,24 @@ var (
...
-					unexportOptions.namespace = "default"
-				}
-			}
-
-			client, err := mcsclient.NewForConfig(config.Config)
-			exit.OnError(status.Error(err, "Error creating client"))
+			exit.OnError(unexportRestConfigProducer.RunOnSelectedContext(
+				func(clusterInfo *cluster.Info, namespace string, status reporter.Interface) error {
+					mcsClient, err := mcsclient.NewForConfig(clusterInfo.RestConfig)
+					exit.OnError(status.Error(err, "Error creating client"))
```
mkolesnik: Do we need `exit.OnError` here?

In other places you removed it

internal/restconfig/restconfig.go
mkolesnik: Thanks, it's usually better (i.e. makes the code more maintainable) to address the exceptional case in the if, not the normal flow, so I'd expect it to look something like this:
```suggestion
	// If we get here, no context was set up, which means SetupFlags() wasn't called
	if rcp.defaultClientConfig == nil {
		return status.Error(errors.New("no context provided (this is a programming error)"), "")
	}
```

internal/restconfig/restconfig.go
```diff
@@ -316,36 +389,34 @@ func (rcp *Producer) ClientConfig() clientcmd.ClientConfig {
...
-
-	if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
-		return nil
-	}
+		if apierrors.IsNotFound(err) || meta.IsNoMatchError(err) {
+			return nil
+		}
 
-	exit.OnErrorWithMessage(err, fmt.Sprintf("Error retrieving Submariner object %s", names.SubmarinerCrName))
+		exit.OnErrorWithMessage(err, fmt.Sprintf("Error retrieving Submariner object %s", names.SubmarinerCrName))
```
mkolesnik: Is this left here intentionally or should it return the error?

cmd/subctl/join.go
mkolesnik: Hmm too bad the broker still needs to be accessed separately. complicating the extraction somewhat. but it's still less unruly :)

mkolesnik: Theoretically you could get an ellipsis or args as optional last argument to the `RunOnSelectedContext` function and pass it on to the nested function, but it might only serve this one use case, while complicating the function signature for everyone :thinking:

internal/restconfig/restconfig.go
skitt: It’s left intentionally for now, this function is used for old flag-handling and new flag-handling.

cmd/subctl/join.go
skitt: Passing `args` would also be useful in `export` and `unexport`, but I’m not sure it’s worth it.

dfarrell07 (approved):

cmd/subctl/azure.go
```diff
@@ -16,28 +16,36 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// nolint:dupl // Similar code in aws.go, azure.go, rhos.go, but not duplicate
```
mkolesnik, line 19: Just a question, why not do the flag init once for all `cloud` commands? They all need the same set of flags.

mkolesnik (approved):

cmd/subctl/azure.go
skitt: Fair question, I’ve updated the PR to share the setup. Thanks!

PR #248, opened 2022-09-01
Don't wait for the timeout on successful firewall tests

### LINKED ISSUE #247: Firewall tests always wait for the timeout
`subctl diagnose` firewall tests always wait for the timeout; they should stop as soon as the test traffic is received.

skitt:
timeout ... tcpdump always waits for the timeout to expire, including
when the test message is received successfully. This patch reworks the
shell command so that it exits as soon as the message is received.
This reduces typical default test times from 90s to 20s.

Fixes: #247

### DIFF
--- a/pkg/diagnose/firewall.go
+++ b/pkg/diagnose/firewall.go
@@ -231,8 +231,11 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, optio
 	}
 
 	clientMessage := string(uuid.NewUUID())[0:8]
-	podCommand := fmt.Sprintf("timeout %d tcpdump -ln -Q in -A -s 100 -i any udp and %s | grep '%s'",
-		options.ValidationTimeout, portFilter, clientMessage)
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

pkg/diagnose/firewall.go
```diff
@@ -231,8 +231,11 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, optio
 	}
 
 	clientMessage := string(uuid.NewUUID())[0:8]
-	podCommand := fmt.Sprintf("timeout %d tcpdump -ln -Q in -A -s 100 -i any udp and %s | grep '%s'",
-		options.ValidationTimeout, portFilter, clientMessage)
+	// The following construct ensures that tcpdump will be stopped as soon as the message is seen, instead of waiting
```
dfarrell07, line 234: Thanks for adding the inline docs!

dfarrell07 (approved):

PR #258, opened 2022-09-07
Set privileged scc using clusterrole and clusterrolebinding

### LINKED ISSUE #117: Subctl join fails (error updating the privileged SCC) on ROSA cluster
**What happened**:

I tried to deploy Submariner on ROSA cluster using subctl : 

subctl join --kubecontext <xx>  broker-info.subm  --clusterid aws-rosa  --load-balancer

Deployment failed with the following error : 

Join error :

✗ Deploying the Submariner operator 
 ✓ Created operator CRDs
 ✓ Created operator namespace: submariner-operator
 ✓ Created operator service account and role
Error deploying the operator: error updating the SCC resource: error updating OpenShift privileged SCC: admission webhook "scc-validation.managed.openshift.io" denied the request: Modifying default SCCs [anyuid hostaccess hostmount-anyuid hostnetwork node-exporter nonroot privileged restricted] is not allowed

subctl version: v0.12.1

Successful deployment 

see above 

As part of Submariner deployment we update [1]  the privileged SCC users with the Submariner Service Accounts , and in ROSA clusters seems that the admission webhook controller is rejecting any changes to the privileged SCC

I was able to  add user to privileged SCC using [2] command

[1] 
https://github.com/submariner-io/subctl/blob/devel/pkg/scc/update.go#L40 

[2]
oc adm policy add-scc-to-user privileged system:serviceaccount:submariner-operator:submariner-gateway

**Environment**:
- Install tools:
- Others:

yboaron:
In current implementation, the relevant submariner components obtain privileged permission by manipulating the system privileged scc CR.

Manipulating the system privileged scc is not allowed in ROSA and ARO platforms, check [1] for more details.

This PR defines privileged permission in OCP platform for the relevant components using ClusterRole and ClusterRoleBinding resources as described in [2] instead of manipulating the  system privileged scc CR.

Depend On: https://github.com/submariner-io/submariner-operator/pull/2237
Fixes: https://github.com/submariner-io/subctl/issues/117

[1]
https://github.com/openshift/managed-cluster-validating-webhooks/pull/176

[2]
https://developer.ibm.com/learningpaths/secure-context-constraints-openshift/scc-tutorial/

Signed-off-by: Yossi Boaron <yboaron@redhat.com>


pkg/scc/update.go
```diff
@@ -19,54 +19,184 @@ limitations under the License.
...
+  name: submariner-privileged-scc
+subjects:
+- kind: ServiceAccount
+  name: submariner-lighthouse-coredns
+  namespace: submariner-operator
+`
+)
+
+var (
+	privilegedClusterRoleDefined = false
```
tpantelis: Why is this defined globally? We really should avoid mutable global state.

pkg/scc/update.go
```diff
@@ -19,54 +19,184 @@ limitations under the License.
...
+
+	nameToYaml = map[string]string{
```
tpantelis: Once the yamls are added to the **submariner-operator** we can have the caller pass in the yaml name instead of using a hardcoded map.

tpantelis (changes requested):

pkg/scc/update.go
yboaron: Hmm, not sure that I understood your comment,
can you explain how the appropriate CRB YAML file should be retrieved given the component name without using something like map ?

pkg/scc/update.go
yboaron: To avoid calling ensure function if ClusterRole resource is already created, I guess we can delete this check.
I'll update the code

tpantelis: Since the ClusterRole only need be created once, perhaps we can let the caller do it in the appropriate place.

pkg/scc/update.go
tpantelis: You mentioned in the PR you defined the "YAML files just to get your feedback and to simplify testing, it should move to submariner-operator repo as the rest of the embedded YAML files.". Once we do that then the caller can just pass in the desired yaml variable name directly, eg `Config_submariner_gateway_privilged_scc_clusterrolebinding_yaml`, instead of maintaining a hardcoded mapping to get it indirectly. We do that in other places.

tpantelis (changes requested):

yboaron:
The build gate is failing because this PR depends on submariner-operator changes that were merged to devel while subctl go.mod points to v0.14.0-m1

skitt:
> The build gate is failing because this PR depends on submariner-operator changes that were merged to devel while subctl go.mod points to v0.14.0-m1

You need to update the dependency: `go get github.com/submariner-io/submariner-operator@devel` (and add the resulting changes to your commit).

sridhargaddam (approved):
LGTM, but I'll let @skitt @tpantelis to comment as they are more familiar with this code.

pkg/operator/ocp/ensure.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+)
+
+func EnsureRBAC(dynClient dynamic.Interface, kubeClient kubernetes.Interface, namespace string) (bool, error) {
+	if !IsOcpPlatform(dynClient) {
+		return false, nil
+	}
+
+	createdCR, err := clusterrole.EnsureFromYAML(kubeClient, embeddedyamls.Config_rbac_submariner_operator_cluster_role_yaml)
```
skitt: How about using the same structure as in `submariner/ocp`, with a data structure describing the files to apply? That way the `EnsureRBAC` function could be shared, passing in the appropriate list of `rbacInfo`s.

pkg/operator/ocp/ensure.go
```diff
@@ -0,0 +1,58 @@
...
+		return false, errors.Wrap(err, "error provisioning the operator OCP ClusterRole resource")
+	}
+
+	createdCRB, err := clusterrolebinding.EnsureFromYAML(kubeClient, namespace,
+		embeddedyamls.Config_rbac_submariner_operator_ocp_cluster_role_binding_yaml)
+	if err != nil {
+		return false, errors.Wrap(err, "error provisioning the operator OCP ClusterRoleBinding resource")
+	}
+
+	return createdCR || createdCRB, errors.Wrap(err, "error updating the SCC resources")
```
skitt: Here the booleans are ored, whereas in `submariner/ocp` they are anded — which should it be?

skitt (changes requested):

pkg/operator/ocp/ensure.go
yboaron: Sounds good to me, I'll update the PR.
10x

pkg/operator/ocp/ensure.go
yboaron: It should be and

pkg/lighthouse/ocp/ensure.go
```diff
@@ -0,0 +1,59 @@
...
+		return false, errors.Wrap(err, "error provisioning the CoreDNS OCP ClusterRole resource")
+	}
+
+	createdCoreDNSCRB, err := clusterrolebinding.EnsureFromYAML(kubeClient, namespace,
+		embeddedyamls.Config_rbac_lighthouse_coredns_cluster_role_binding_yaml)
+	if err != nil {
+		return false, errors.Wrap(err, "error provisioning the CoreDNS OCP ClusterRoleBinding resource")
+	}
+
+	return createdAgentCR || createdAgentCRB || createdCoreDNSCR || createdCoreDNSCRB, errors.Wrap(err, "error updating the SCC resources")
```
tpantelis: `err` will be nil here since it's checked on L54.

```suggestion
	return createdAgentCR || createdAgentCRB || createdCoreDNSCR || createdCoreDNSCRB, nil
```

pkg/lighthouse/ensure.go
```diff
@@ -35,7 +35,7 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
 		status.Success("Created lighthouse service account and role")
 	}
 
-	if created, err := scc.Ensure(dynClient, operatorNamespace); err != nil {
+	if created, err := ocp.EnsureRBAC(dynClient, kubeClient, operatorNamespace); err != nil {
 		return err // nolint:wrapcheck // No need to wrap here
 	} else if created {
 		status.Success("Updated the privileged SCC")
```
tpantelis: Should this change (elsewhere as well)?

```suggestion
		status.Success("Updated the privileged OCP roles")
```

pkg/operator/ocp/ensure.go
tpantelis: They should be OR'ed - the intent is to return true if any were created.

Also `err` will be nil here.

pkg/operator/ocp/ensure.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func IsOcpPlatform(dynClient dynamic.Interface) bool {
+	sccClient := dynClient.Resource(securityv1.GroupVersion.WithResource("securitycontextconstraints"))
+	_, err := sccClient.Get(context.TODO(), "privileged", metav1.GetOptions{})
+
+	return err == nil
```
tpantelis, line 67: Technically we really should check if it's a `NotFound` error and fail on any other error but I guess this is fine - if we can't access the API server then something else will likely fail.

pkg/submariner/ocp/ensure.go
```diff
@@ -0,0 +1,88 @@
...
+			clusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_diagnose_ocp_cluster_role_binding_yaml,
+		},
+		{
+			componentName:          "networkplugin-syncer",
+			clusterRoleFile:        embeddedyamls.Config_rbac_networkplugin_syncer_ocp_cluster_role_yaml,
+			clusterRoleBindingFile: embeddedyamls.Config_rbac_networkplugin_syncer_ocp_cluster_role_binding_yaml,
+		},
+	}
+
+	updateRbac := true
```
tpantelis:
```suggestion
	updatedRbac := false
```

pkg/submariner/ocp/ensure.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			return false, errors.Wrapf(err, "error provisioning the %s OCP ClusterRole resource", componentRbac.componentName)
+		}
+
+		createdCRB, err := clusterrolebinding.EnsureFromYAML(kubeClient, namespace, componentRbac.clusterRoleBindingFile)
+		if err != nil {
+			return false, errors.Wrapf(err, "error provisioning the %s OCP ClusterRoleBinding resource", componentRbac.componentName)
+		}
+
+		updateRbac = updateRbac && createdCR && createdCRB
```
tpantelis:
```suggestion
		updatedRbac = updatedRbac || createdCR || createdCRB
```

tpantelis (changes requested):

pkg/operator/ocp/ensure.go
yboaron: Hmm, but even if one of clusterrole , clusterrolebinding was created, we won't have privileged permission for the component, what am I missing here?

tpantelis: The purpose of the return boolean is to report if something was actually created (successfully), ie if it didn't already exist, so the caller can report that to the user. A resource may already exist if `join` was previously run. The return error indicates if a creation failed.

yboaron: Ack, 10x. will change to or

pkg/lighthouse/ensure.go
```diff
@@ -35,10 +36,23 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_lighthouse_coredns_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_lighthouse_coredns_cluster_role_binding_yaml,
+		},
+	}
+
+	if created, err := ocp.EnsureRBAC(dynClient, kubeClient, operatorNamespace, componentsRbac); err != nil {
 		return err // nolint:wrapcheck // No need to wrap here
 	} else if created {
-		status.Success("Updated the privileged SCC")
+		status.Success("Updated the OCP rolse")
```
skitt:
```suggestion
		status.Success("Updated the OCP roles")
```

pkg/operator/ensure.go
```diff
@@ -56,10 +57,19 @@ func Ensure(status reporter.Interface, clientProducer client.Producer, operatorN
...
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_operator_ocp_cluster_role_binding_yaml,
+		},
+	}
+
+	if created, err := ocp.EnsureRBAC(clientProducer.ForDynamic(), clientProducer.ForKubernetes(),
+		operatorNamespace, componentsRbac); err != nil {
 		return err
 	} else if created {
-		status.Success("Updated the privileged SCC")
+		status.Success("Updated the OCP rolse")
```
skitt:
```suggestion
		status.Success("Updated the OCP roles")
```

pkg/submariner/ensure.go
```diff
@@ -35,10 +36,38 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_networkplugin_syncer_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_networkplugin_syncer_ocp_cluster_role_binding_yaml,
+		},
+	}
+
+	if created, err := ocp.EnsureRBAC(dynClient, kubeClient, operatorNamespace, componentsRbac); err != nil {
 		return err // nolint:wrapcheck // No need to wrap here
 	} else if created {
-		status.Success("Updated the privileged SCC")
+		status.Success("Updated the OCP rolse")
```
skitt:
```suggestion
		status.Success("Updated the OCP roles")
```

skitt (changes requested):
Thanks for this, there’s a small typo to fix if you don’t mind.

pkg/lighthouse/ensure.go
```diff
@@ -35,10 +36,23 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
 		status.Success("Created lighthouse service account and role")
 	}
 
-	if created, err := scc.Ensure(dynClient, operatorNamespace); err != nil {
+	componentsRbac := []ocp.RbacInfo{
+		{
+			ComponentName:          "lighthouse-agent",
```
tpantelis:
```suggestion
			ComponentName:          names.ServiceDiscoveryComponent,
```

pkg/lighthouse/ensure.go
```diff
@@ -35,10 +36,23 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_lighthouse_agent_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_lighthouse_agent_ocp_cluster_role_binding_yaml,
+		},
+		{
+			ComponentName:          "lighthouse-coredns",
```
tpantelis:
```suggestion
			ComponentName:          names.LighthouseCoreDNSComponent,
```

pkg/operator/ensure.go
```diff
@@ -56,10 +57,19 @@ func Ensure(status reporter.Interface, clientProducer client.Producer, operatorN
 		status.Success("Created operator service account and role")
 	}
 
-	if created, err := scc.Ensure(clientProducer.ForDynamic(), operatorNamespace); err != nil {
+	componentsRbac := []ocp.RbacInfo{
+		{
+			ComponentName:          "submariner-operator",
```
tpantelis:
```suggestion
			ComponentName:          names.OperatorComponent,
```

pkg/submariner/ensure.go
```diff
@@ -35,10 +36,38 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
 		status.Success("Created submariner service account and role")
 	}
 
-	if created, err := scc.Ensure(dynClient, operatorNamespace); err != nil {
+	componentsRbac := []ocp.RbacInfo{
+		{
+			ComponentName:          "submariner-gateway",
```
tpantelis:
```suggestion
			ComponentName:          names.GatewayComponent,
```

pkg/submariner/ensure.go
```diff
@@ -35,10 +36,38 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_submariner_gateway_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_gateway_ocp_cluster_role_binding_yaml,
+		},
+		{
+			ComponentName:          "submariner-route-agent",
```
tpantelis:
```suggestion
			ComponentName:          names.RouteAgentComponent,
```

pkg/submariner/ensure.go
```diff
@@ -35,10 +36,38 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_submariner_route_agent_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_route_agent_ocp_cluster_role_binding_yaml,
+		},
+		{
+			ComponentName:          "submariner-globalnet",
```
tpantelis:
```suggestion
			ComponentName:          names.GlobalnetComponent,
```

pkg/submariner/ensure.go
```diff
@@ -35,10 +36,38 @@ func Ensure(status reporter.Interface, kubeClient kubernetes.Interface, dynClien
...
+			ClusterRoleFile:        embeddedyamls.Config_rbac_submariner_globalnet_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_globalnet_ocp_cluster_role_binding_yaml,
+		},
+		{
+			ComponentName:          "submariner-diagnose",
+			ClusterRoleFile:        embeddedyamls.Config_rbac_submariner_diagnose_ocp_cluster_role_yaml,
+			ClusterRoleBindingFile: embeddedyamls.Config_rbac_submariner_diagnose_ocp_cluster_role_binding_yaml,
+		},
+		{
+			ComponentName:          "networkplugin-syncer",
```
tpantelis:
```suggestion
			ComponentName:          names.NetworkPluginSyncerComponent,
```

tpantelis (changes requested):

tpantelis (approved):

skitt:
https://github.com/submariner-io/subctl/pull/282 should fix the system tests.

skitt (approved):

PR #259, opened 2022-09-08
Don't specify images to preload

mkolesnik:
Shipyard now preloads all built images,
there's no need to explicitly preload any images.

Depends on https://github.com/submariner-io/shipyard/pull/937

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,6 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 PLATFORMS ?= linux/amd64,linux/arm64
 IMAGES = subctl
-PRELOAD_IMAGES := submariner-operator submariner-gateway submariner-globalnet submariner-route-agent lighthouse-agent lighthouse-coredns
 MULTIARCH_IMAGES := subctl
 
 ifneq (,$(filter ovn,$(USING)))


Makefile
```diff
@@ -24,7 +24,6 @@ ifneq (,$(DAPPER_HOST_ARCH))
 
 PLATFORMS ?= linux/amd64,linux/arm64
 IMAGES = subctl
-PRELOAD_IMAGES := submariner-operator submariner-gateway submariner-globalnet submariner-route-agent lighthouse-agent lighthouse-coredns
```
tpantelis, line 27: FWICT, the changes in shipyard will now preload $IMAGES which would now only be `subctl`, ie the others specified here won't be preloaded or am I missing something?

mkolesnik: Yes that's precisely the point, each project tests it's own built images by default, and as a developer you can explicitly preload extra images when you need to (which would be quite rare AFAICT).

Nowadays everything is preloaded from the local docker registry which, most often than not, contains stale images which lead to buggy and unexpected behaviors.

tpantelis (approved):

dfarrell07 (approved):

PR #264, opened 2022-09-13
Check if metrics-proxy daemonset is deployed

vthapar:

### DIFF
--- a/pkg/diagnose/deployments.go
+++ b/pkg/diagnose/deployments.go
@@ -129,6 +129,8 @@ func checkPods(clusterInfo *cluster.Info, status reporter.Interface) bool {
 		checkDeployment(clusterInfo.ClientProducer.ForKubernetes(), constants.OperatorNamespace, "submariner-networkplugin-syncer", tracker)
 	}
 
+	checkDaemonset(clusterInfo.ClientProducer.ForKubernetes(), clusterInfo.Submariner.Namespace, "submariner-metrics-proxy", tracker)
+
 	checkPodsStatus(clusterInfo.ClientProducer.ForKubernetes(), constants.OperatorNamespace, tracker)
 
 	if tracker.HasFailures() {


tpantelis (approved):

pkg/diagnose/deployments.go
```diff
@@ -129,6 +129,8 @@ func checkPods(clusterInfo *cluster.Info, status reporter.Interface) bool {
 		checkDeployment(clusterInfo.ClientProducer.ForKubernetes(), constants.OperatorNamespace, "submariner-networkplugin-syncer", tracker)
 	}
 
+	checkDaemonset(clusterInfo.ClientProducer.ForKubernetes(), clusterInfo.Submariner.Spec.Namespace, "submariner-metrics-proxy", tracker)
```
tpantelis: `Submariner.Spec.Namespace` is used for the `SUBMARINER_NAMESPACE` env var passed to the submariner pods. It is set from `constants.SubmarinerNamespace`. `constants.OperatorNamespace` is where the `Submariner` resource and thus all the DaemonSets, Deployments et al are installed. So semantically they're different although they're set to the same value. However we seem to use them interchangeably in several places. So technically `constants.OperatorNamespace` is the correct usage here. Better yet, to avoid using a hardcoded constant, we should use `Submariner.Namespace`.  

We really should use `Submariner.Namespace` in place of `constants.OperatorNamespace` in all places in this function to be consistent. We could do that in a separate PR but I'm fine either way.

tpantelis (changes requested):

pkg/diagnose/deployments.go
vthapar: I'll raise a separate PR to change constant to `Submariner.Namespace`

yboaron (approved):

tpantelis (approved):

pkg/diagnose/deployments.go
tpantelis: As a follow-up to https://github.com/submariner-io/subctl/pull/265, I'm going to reduce usages of `constants.OperatorNamespace` as well so I can make the changes here.

PR #306, opened 2022-10-18
Wrap two-context diagnose commands

skitt:
This updates diagnose firewall inter-cluster and nat-discovery to use the new restconfig wrappers. They use specific local/remote producers for the "standard" scenario (--context <localcontext> --remotecontext <remotecontext>) and a legacy producer, not connected to flags, to handle the deprecated two-kubeconfig argument form.

The two Cobra runners end up being very similar, so they are moved into a separate runLocalRemoteCommand() wrapper. For both commands, the namespace flag is taken care of by restconfig, which means we need to move namespace flag handling out and take care of it separately for the other namespaced commands which aren't migrated in this commit.

System tests are added for the new variants, including kubeconfig variants which aren't yet supported (commented out, but preserved to reduce the risk that we forget about them).

### DIFF
--- a/cmd/subctl/diagnose.go
+++ b/cmd/subctl/diagnose.go
@@ -41,6 +41,11 @@ var (
 
 	diagnoseRestConfigProducer = restconfig.NewProducer()
 
+	diagnoseFirewallTunnelRestConfigProducer = restconfig.NewProducer().
+							WithDefaultNamespace(constants.OperatorNamespace).WithPrefixedContext("remote")
+	diagnoseFirewallNatDiscoveryRestConfigProducer = restconfig.NewProducer().
+							WithDefaultNamespace(constants.OperatorNamespace).WithPrefixedContext("remote")
+
 	diagnoseCmd = &cobra.Command{
 		Use:   "diagnose",
 		Short: "Run diagnostic checks on the Submariner deployment and report any issues",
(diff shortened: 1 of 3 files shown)


dfarrell07 (approved):

cmd/subctl/diagnose.go
```diff
@@ -41,6 +41,11 @@ var (
 
 	diagnoseRestConfigProducer = restconfig.NewProducer()
 
+	diagnoseFirewallTunnelRestConfigProducer = restconfig.NewProducer().
+							WithDefaultNamespace(constants.OperatorNamespace).WithPrefixedContext("remote")
+	diagnoseFirewallNatDiscoveryRestConfigProducer = restconfig.NewProducer().
+							WithDefaultNamespace(constants.OperatorNamespace).WithPrefixedContext("remote")
```
mkolesnik, line 47: Wouldn't it be simpler to store it in one producer?

e.g.
```suggestion
	diagnoseFirewallRemoteConfigProducer = restconfig.NewProducer().
							WithDefaultNamespace(constants.OperatorNamespace).WithPrefixedContext("remote")
```

cmd/subctl/diagnose.go
```diff
@@ -211,41 +213,6 @@ func addDiagnoseFWConfigFlags(command *cobra.Command) {
 		"time to run in seconds while validating the firewall")
 	command.Flags().BoolVar(&diagnoseFirewallOptions.VerboseOutput, "verbose", false,
 		"produce verbose output while validating the firewall")
-	addDiagnosePodNamespaceFlag(command, &diagnoseFirewallOptions.PodNamespace)
```
mkolesnik, line 214: I didn't understand why this call was moved out of this function, it just seems to add code duplication.

cmd/subctl/diagnose.go
skitt: It would, but it doesn’t work — a given variable can only be used with a single command’s flags. There’s no shared parent command that could be used with persistent flags either.

cmd/subctl/diagnose.go
skitt: This function is called on all `firewall` sub-commands, but `inter-cluster` and `nat-discovery` now use the shared namespace support, so we can’t add a specific namespace flag there. As a result, the `addDiagnosePodNamespaceFlag` call must only be made on the other sub-commands.

mkolesnik: OK thanks for explaining, its getting a bit difficult to track all these changes :)

mkolesnik (approved):

dfarrell07:
[Subctl Commands (globalnet lighthouse)](https://github.com/submariner-io/subctl/actions/runs/3282316942/jobs/5405554613#logs) failed, re-running.

tpantelis (approved):

PR #312, opened 2022-10-20
Further migration of kubeconfig handling

### LINKED ISSUE #196: `subctl diagnose` and `subctl gather` don't return error exit code on empty/non-existant `KUBECONFIG`
**What happened**:
This might go deeper than this (e.g other error situations), but just running these commands with no `KUBECONFIG` set results in exit code 0, but the output indicates error:
```
No kube config was provided. Please use the --kubeconfig flag or set the KUBECONFIG environment variable
```

Also, the output gets written to `stdout` while it should go to `stderr`.

I expect some exit code other than **0**.

```bash
KUBECONFIG= subctl diagnose all
KUBECONFIG= subctl gather all
```

The same happens if KUBECONFIG points to a non existing file, but not when using `--kubeconfig` flag (then is behaves as expected).

**Environment**:
- Install tools:
- Others:

skitt:
This completes `subctl diagnose` and adds `subctl show` and `subctl gather`. See individual commits for details.

Fixes: #196


dfarrell07 (approved):

pkg/diagnose/connections.go
```diff
@@ -76,11 +75,9 @@ func Connections(clusterInfo *cluster.Info, status reporter.Interface) bool {
...
-	if tracker.HasFailures() {
-		return false
+	if !tracker.HasFailures() {
+		status.Success("All connections are established")
 	}
 
-	status.Success("All connections are established")
-
-	return true
+	return nil
```
tpantelis, line 84: Shouldn't this return an error to fail the command?

cmd/subctl/diagnose.go
```diff
@@ -215,54 +207,39 @@ func addDiagnoseFWConfigFlags(command *cobra.Command) {
...
-	fmt.Println()
+var allDiagnoseCommands = []restconfig.PerContextFn{
+	diagnose.K8sVersion,
+	restconfig.IfSubmarinerInstalled(diagnose.CNIConfig),
+	restconfig.IfSubmarinerInstalled(diagnose.Connections),
+	restconfig.IfSubmarinerInstalled(diagnose.Deployments),
+	restconfig.IfSubmarinerInstalled(diagnose.KubeProxyMode),
+	restconfig.IfSubmarinerInstalled(firewallIntraVxLANConfig),
+	restconfig.IfSubmarinerInstalled(diagnose.GlobalnetConfig),
+	restconfig.IfServiceDiscoveryInstalled(diagnose.ServiceDiscovery),
```
tpantelis, line 224: So because each sub-command now checks `IfSubmarinerInstalled`  we get this output:

```
Cluster "cluster1"
 ✓ Checking Submariner support for the Kubernetes version
 ✓ Kubernetes version "v1.24.3" is supported

 ⚠ Submariner is not installed

 ⚠ Submariner is not installed

 ⚠ Submariner is not installed

 ⚠ Submariner is not installed

 ⚠ Submariner is not installed

 ⚠ Submariner is not installed

 ⚠ Service discovery is not installed
```

Which looks a little awkward. Perhaps it would be clearer if we could print what's being skipped, ie:

`⚠ Skipping check for Submariner CNI network plugin support - Submariner is not installed`

We could pass an optional context string to `IfSubmarinerInstalled`:

```Go
func IfSubmarinerInstalled(context string, run PerContextFn) PerContextFn {
	return func(clusterInfo *cluster.Info, namespace string, status reporter.Interface) error {
		if clusterInfo.Submariner == nil {
                         if context != "" {
                             status.Warningf("Skipping %s - %s, context, constants.SubmarinerNotInstalled)
                         } else {
			    status.Warning(constants.SubmarinerNotInstalled)
                         }

			return nil
		}

		return run(clusterInfo, namespace, status)
	}
}
```

internal/restconfig/restconfig.go
```diff
@@ -256,47 +319,95 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+	if rcp.inCluster {
+		return runInCluster(function, status)
+	}
+
+	if rcp.defaultClientConfig.overrides.CurrentContext != "" {
+		// The user has explicitly chosen a context, use that only
+		return rcp.RunOnSelectedContext(function, status)
+	}
+
+	if rcp.defaultClientConfig != nil {
```
tpantelis: To reduce indentation:

```suggestion
	if rcp.defaultClientConfig == nil {
	    return nil
	}
```

internal/restconfig/restconfig.go
```diff
@@ -256,47 +319,95 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+			for _, contextName := range rcp.contexts {
+				chosenContext, ok := rawConfig.Contexts[contextName]
+				if !ok {
+					status.Failure("error retrieving context %s", contextName)
+
+					continue
+				}
+
+				if err := rcp.overrideContextAndRun(chosenContext.Cluster, contextName, function, status); err != nil {
+					return err
```
tpantelis: So if one cluster fails it will skip the rest which is a change from before.

pkg/diagnose/cni.go
```diff
@@ -167,7 +163,11 @@ func checkCalicoIPPoolsIfCalicoCNI(info *cluster.Info, status reporter.Interface
 
 	checkGatewaySubnets(gateways, ippools, tracker)
 
-	return !tracker.HasFailures()
+	if tracker.HasFailures() {
+		return errors.New("Failures while diagnosing CNI")
```
tpantelis: Why didn't linting flag this...

```suggestion
		return errors.New("failures while diagnosing CNI")
```

pkg/diagnose/deployments.go
```diff
@@ -92,7 +100,7 @@ func checkOverlappingCIDRs(clusterInfo *cluster.Info, status reporter.Interface)
 	}
 
 	if tracker.HasFailures() {
-		return false
+		return errors.New("Failures while diagnosing overlapping CIDRs")
```
tpantelis:
```suggestion
		return errors.New("failures while diagnosing overlapping CIDRs")
```

pkg/diagnose/deployments.go
```diff
@@ -134,12 +142,12 @@ func checkPods(clusterInfo *cluster.Info, status reporter.Interface) bool {
 	checkPodsStatus(clusterInfo.ClientProducer.ForKubernetes(), constants.OperatorNamespace, tracker)
 
 	if tracker.HasFailures() {
-		return false
+		return errors.New("Failures while diagnosing pods")
```
tpantelis:
```suggestion
		return errors.New("failures while diagnosing pods")
```

pkg/diagnose/firewall_vxlan.go
```diff
@@ -30,34 +31,33 @@ const (
...
+		return nil
 	}
 
 	tracker := reporter.NewTracker(status)
 
 	checkFWConfig(clusterInfo, options, tracker)
 
 	if tracker.HasFailures() {
-		return false
+		return errors.New("Failures while diagnosing the intra-VXLAN firewall configuration")
```
tpantelis:
```suggestion
		return errors.New("failures while diagnosing the intra-VXLAN firewall configuration")
```

pkg/diagnose/k8s_version.go
```diff
@@ -41,10 +41,10 @@ func K8sVersion(client kubernetes.Interface, status reporter.Interface) bool {
 	}
 
 	if tracker.HasFailures() {
-		return false
+		return errors.New("Failures while diagnosing the Kubernetes version")
```
tpantelis:
```suggestion
		return errors.New("failures while diagnosing the Kubernetes version")
```

pkg/diagnose/cni.go
```diff
@@ -107,44 +107,40 @@ func detectCalicoConfigMap(clientSet kubernetes.Interface) (bool, error) {
...
 	if err != nil {
-		status.Failure("Error retrieving Gateways: %v", err)
-		return false
+		return status.Error(err, "Error retrieving Gateways")
 	}
 
 	if len(gateways) == 0 {
 		status.Warning("There are no gateways detected on the cluster")
-		return false
+		return errors.New("no gateways on the cluster")
```
tpantelis: It's a little odd to log a warning yet fail the command - perhaps we should use `return status.Error(err, "No gateways detected on the cluster"`).

pkg/diagnose/cni.go
```diff
@@ -73,9 +73,9 @@ func CNIConfig(clusterInfo *cluster.Info, status reporter.Interface) bool {
 	}
 
 	if !isSupportedPlugin {
-		status.Failure("The detected CNI network plugin (%q) is not supported by Submariner."+
-			" Supported network plugins: %v\n", clusterInfo.Submariner.Status.NetworkPlugin, supportedNetworkPlugins)
-		return false
+		return status.Error(errors.New("Unsupported CNI plugin"),
+			"The detected CNI network plugin (%q) is not supported by Submariner."+
+				" Supported network plugins: %v\n", clusterInfo.Submariner.Status.NetworkPlugin, supportedNetworkPlugins)
```
tpantelis: The detailed info is in the passed message so appending the error string is redundant and looks a bit awkward.

```suggestion
		status.Failure("The detected CNI network plugin (%q) is not supported by Submariner."+
				" Supported network plugins: %v\n", clusterInfo.Submariner.Status.NetworkPlugin, supportedNetworkPlugins)
		return errors.New("unsupported CNI plugin")
```

pkg/diagnose/cni.go
```diff
@@ -224,32 +224,30 @@ func mustHaveSubmariner(clusterInfo *cluster.Info) {
...
-		status.Failure("Failed to get ovn-nb database version %v", err)
-		return false
+		return status.Error(err, "Failed to get ovn-nb database version")
 	}
 
 	if version.Compare(ovnNBVersion, minOVNNBVersion, "<") {
-		status.Failure("The ovn-nb database version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
-		return false
+		return status.Error(errors.New("Unsupported ovn-db database version"),
+			"The ovn-nb database version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
```
tpantelis:
```suggestion
                 status.Failure("The ovn-nb database version %v is less than the minimum supported version %v", ovnNBVersion, minOVNNBVersion)
		return errors.New("unsupported ovn-db database version")
```

pkg/diagnose/deployment_metrics.go
```diff
@@ -31,53 +33,54 @@ const (
...
-		return false
+		return status.Error(err, "Error waiting for the client pod to finish its execution")
 	}
 
 	// Expected response: "HTTP/1.1 200 OK"
 	if !strings.Contains(cPod.PodOutput, "200 OK") {
-		status.Failure("Unable to access %s metrics service in submariner-operator "+
-			"namespace %v", component, cPod.PodOutput)
-		return false
+		return status.Error(errors.Errorf("Unexpected output %v", cPod.PodOutput),
```
tpantelis, line 79:
```suggestion
		return status.Error(errors.Errorf("unexpected output %v", cPod.PodOutput),
```

internal/restconfig/restconfig.go
```diff
@@ -256,47 +319,95 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+			}
+		} else {
+			// Loop over all accessible contexts
+			for contextName, context := range rawConfig.Contexts {
+				if err := rcp.overrideContextAndRun(context.Cluster, contextName, function, status); err != nil {
+					return err
+				}
+			}
+		}
+	}
```
tpantelis, line 374: If there's no kube configs/contexts it now silently returns while, previously, `OnMultiCluster` printed a message. It also doesn't fail (see https://github.com/submariner-io/subctl/issues/196). Is this intentional?

tpantelis (changes requested):

cmd/subctl/diagnose.go
skitt: Or we could warn only once (remembering that we’ve warned). I’m not sure most users care about what tests we’re skipping!

pkg/diagnose/connections.go
skitt: It should indeed.

cmd/subctl/diagnose.go
tpantelis: yeah that's what it did before - either way is fine with me.

internal/restconfig/restconfig.go
skitt: Indeed, fixed.

internal/restconfig/restconfig.go
skitt: Right, I meant to address that later but I should do it now.

skitt:
Hmm, I’ve pushed an update (see https://github.com/skitt/subctl/tree/kubeconfig-subctl-gather) but it’s not updating this PR...

internal/restconfig/restconfig.go
```diff
@@ -547,3 +583,39 @@ func ConfigureTestFramework(args []string) error {
 
 	return nil
 }
+
+var submarinerWarned = false
```
skitt: This means that if we process multiple contexts, we’ll only get a warning for the first one :-/

internal/restconfig/restconfig.go
```diff
@@ -256,121 +320,109 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+	processedContexts := 0
 
-	return configs
-}
+	if len(rcp.contexts) > 0 {
+		// Loop over explicitly-chosen contexts
+		for _, contextName := range rcp.contexts {
+			chosenContext, ok := rawConfig.Contexts[contextName]
+			if !ok {
+				status.Failure("error retrieving context %s", contextName)
```
tpantelis, line 357: Shouldn't we add a `contextError` here? Otherwise it seems it will log a failure message but won't fail it.

internal/restconfig/restconfig.go
```diff
@@ -256,121 +320,109 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+			processedContexts++
 
-	if len(restConfigs) > 0 {
-		return restConfigs[0], nil
+			contextErrors = append(contextErrors, rcp.overrideContextAndRun(context.Cluster, contextName, function, status))
+		}
 	}
 
-	return restConfig, errors.New("error getting restconfig")
+	return processedContexts > 0, k8serrors.NewAggregate(contextErrors)
```
tpantelis: So all the call sites want to `failIfNoContext` - I don't see a case where we wouldn't - so I suggest we just handle it here instead of returning a boolean and requiring callers to do it.

pkg/diagnose/cni.go
```diff
@@ -107,44 +107,39 @@ func detectCalicoConfigMap(clientSet kubernetes.Interface) (bool, error) {
...
 	if err != nil {
-		status.Failure("Error retrieving Gateways: %v", err)
-		return false
+		return status.Error(err, "Error retrieving Gateways")
 	}
 
 	if len(gateways) == 0 {
-		status.Warning("There are no gateways detected on the cluster")
-		return false
+		return status.Error(errors.New("no gateways on the cluster"), "No gateways detected on the cluster")
```
tpantelis:
```suggestion
		return status.Error(errors.New("no gateways detected on the cluster"), "")
```

tpantelis (changes requested):

cmd/subctl/diagnose.go
```diff
@@ -215,54 +208,49 @@ func addDiagnoseFWConfigFlags(command *cobra.Command) {
...
-
-	fmt.Println()
-
-	if clusterInfo.Submariner == nil {
-		status.Warning(constants.SubmarinerNotInstalled)
-
-		return success
+func failIfNoContext(contextPresent bool, err error) error {
+	if !contextPresent {
+		return errors.New("no context was specified or found")
```
tpantelis: It seems this will fail the command but not print a message. 

```suggestion
                 fmt.Fprintln(os.Stderr, "No kube config or context was specified or found.")
		return errors.New("no context was specified or found")
```

tpantelis (changes requested):

internal/restconfig/restconfig.go
```diff
@@ -277,6 +291,50 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
+	if rcp.inCluster {
+		return runInCluster(function, status)
+	}
+
+	if rcp.defaultClientConfig.overrides.CurrentContext != "" {
+		// The user has explicitly chosen a context, use that only
+		return rcp.RunOnSelectedContext(function, status)
+	}
+
+	if rcp.defaultClientConfig == nil {
```
tpantelis: defaultClientConfig is accessed on L303 so this check is unnecessary unless it should be before L303

tpantelis (changes requested):

internal/restconfig/restconfig.go
```diff
@@ -256,121 +320,114 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
-		}
+				continue
+			}
 
-		return contexts.Size()
-	}
-	// Current context or rcp.kubeContext
-	return 1
-}
+			processedContexts++
```
tpantelis: This should be before L356 above otherwise we could fail it b/c of no context which would be misleading.

internal/restconfig/restconfig.go
tpantelis: Shouldn't we add a contextError here? Otherwise it seems it will log a failure message but won't fail it.

tpantelis (changes requested):
Looks good - just a couple minor things.

internal/restconfig/restconfig.go
```diff
@@ -256,121 +320,115 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
...
-	if len(rcp.kubeContexts) > 0 {
-		// Count unique contexts
-		contexts := stringset.New()
-		for i := range rcp.kubeContexts {
-			contexts.Add(rcp.kubeContexts[i])
-		}
+			chosenContext, ok := rawConfig.Contexts[contextName]
+			if !ok {
+				contextErrors = append(contextErrors, fmt.Errorf("error retrieving context %s", contextName))
+				status.Failure("error retrieving context %s", contextName)
```
tpantelis:
```suggestion
				contextErrors = append(contextErrors, status.Error(fmt.Errorf("Kubernetes context %q not found", contextName), ""))
```

tpantelis (changes requested):

internal/restconfig/restconfig.go
skitt: Hah, this time the linter flags the leading capital letter!

tpantelis (approved):

PR #314, opened 2022-10-20
Return exit code on empty/non-existent KUBECONFIG

### LINKED ISSUE #196: `subctl diagnose` and `subctl gather` don't return error exit code on empty/non-existant `KUBECONFIG`

tpantelis:
Fixes https://github.com/submariner-io/subctl/issues/196

### DIFF
--- a/cmd/subctl/execute/execute.go
+++ b/cmd/subctl/execute/execute.go
@@ -34,8 +34,8 @@ type OnClusterFn func(*cluster.Info, reporter.Interface) bool
 func OnMultiCluster(restConfigProducer *restconfig.Producer, run OnClusterFn) {
 	restConfigs := restConfigProducer.MustGetForClusters()
 	if len(restConfigs) == 0 {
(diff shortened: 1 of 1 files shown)


dfarrell07 (approved):

skitt:
`OnMultiCluster` is [going away](https://github.com/submariner-io/subctl/pull/312/commits/f92609d163af0d8b5b0441ff6030bff18bc8cbb0).

PR #316, opened 2022-10-23
Use `nettest` image/component names from operator

mkolesnik:
Instead of hard coding the name of the image and the component. Use the ones from the operator.
This will allow downstream forks to override with a different image name if desired.

Depends on https://github.com/submariner-io/submariner-operator/pull/2293

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/submariner-io/lighthouse v0.14.0-m2.0.20221013155253-8a8c1fbfc8d3
 	github.com/submariner-io/shipyard v0.14.0-m2
 	github.com/submariner-io/submariner v0.14.0-m1.0.20220926163523-c1aa0f37d980
-	github.com/submariner-io/submariner-operator v0.14.0-m2.0.20221014112443-3df6539abaf1
+	github.com/submariner-io/submariner-operator v0.14.0-m2.0.20221024122505-adc24314a409
 	github.com/uw-labs/lichen v0.1.7
 	golang.org/x/oauth2 v0.0.0-20221014153046-6fdb5e3db783
 	google.golang.org/api v0.100.0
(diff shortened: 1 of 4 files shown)


nyechiel:
@mkolesnik should we add this as a blocker for 0.14.0-rc0?

mkolesnik:
> @mkolesnik should we add this as a blocker for 0.14.0-rc0?

No, we can always backport it but it shouldn't block the release

nyechiel:
> > @mkolesnik should we add this as a blocker for 0.14.0-rc0?
> 
> No, we can always backport it but it shouldn't block the release

But if we backport later on, would this require another rc and test day? Or are you saying that this is a clean and safe change?

mkolesnik:
> > > @mkolesnik should we add this as a blocker for 0.14.0-rc0?
> > 
> > 
> > No, we can always backport it but it shouldn't block the release
> 
> But if we backport, this would require another rc and test day? Or are you saying this is a clean and safe change?

We're not planning RC1?

This is safe since metrics proxy was just recently added and AFAICT theres no usage of the actual override, so it should be quite safe to backport.

nyechiel:
> > > > @mkolesnik should we add this as a blocker for 0.14.0-rc0?
> > > 
> > > 
> > > No, we can always backport it but it shouldn't block the release
> > 
> > 
> > But if we backport, this would require another rc and test day? Or are you saying this is a clean and safe change?
> 
> We're not planning RC1?
> 

Not at this point (https://github.com/submariner-io/releases/issues/416), but based on feedback (and bugs reported) with rc0 with might need one.

> This is safe since metrics proxy was just recently added and AFAICT theres no usage of the actual override, so it should be quite safe to backport.

tpantelis (approved):

skitt (approved):

mkolesnik:
Backported as https://github.com/submariner-io/subctl/pull/336

PR #332, opened 2022-10-26
Allow --contexts on subctl show

skitt:
This allows users to specify contexts to use, rather than all the accessible contexts.


mkolesnik (approved):

mkolesnik (commented):
I think it would be nice to add a system test, at least one which runs e.g. `--contexts` with a single cluster

tpantelis (approved):

PR #344, opened 2022-10-31
Avoid errors caused by sponge's SIGPIPE

skitt:
In system tests, when the tests produce large amounts of output which is ultimately fed to "grep -q", if the sought-after pattern is found early enough, the large quantity of pending output results in a SIGPIPE in sponge (manifesting itself as a 141 error from Make, 141 being 128 + 13, SIGPIPE).

Using "(sponge ||:)" instead of sponge causes sponge's exit status to effectively be ignored. tee doesn't need this handling, ignoring sponge is sufficient.

### DIFF
--- a/scripts/test/system.sh
+++ b/scripts/test/system.sh
@@ -133,13 +133,13 @@ deploy_env_once
 
 # Test subctl show invocations
 
-_subctl show all | tee /dev/stderr | sponge | grep -q 'Cluster "cluster2"'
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

mkolesnik (approved):
I wonder if this can all be folded into a function (or alias or something else) to avoid the repetitiveness and allow fixing such issues all in one place

skitt:
> I wonder if this can all be folded into a function (or alias or something else) to avoid the repetitiveness and allow fixing such issues all in one place

Good idea, https://github.com/submariner-io/subctl/issues/345

PR #347, opened 2022-10-31
Use golangci-lint for header checks

skitt:
goheader's whitespace handling seems to be less fragile than license_checker's. This also means one less job in CI.


nyechiel:
@skitt should we get this into 0.14? If so, let's add to the rc2 blocker as well.

skitt:
> @skitt should we get this into 0.14? If so, let's add to the rc2 blocker as well.

No, this isn’t relevant for 0.14.

tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
> License Headers Expected — Waiting for status to be reported Required

Existing job needs to be made not-required before this can be merged.

Jaanki (approved):

Issue #348, opened 2022-10-31
OCP 4.12: `subctl diagnose` shows pod security warnings


nyechiel:
- `subctl`-based deployment on top of OCP 4.12 (k8s 1.25) clusters using `0.14.0-rc1`
- Installation seems to be working fine; IPsec connections are up
- `subctl diagnose all` shows the following warnings:

```
$ subctl diagnose all
Cluster "nyechiel-ovnk-2"
 ✓ Checking Submariner support for the Kubernetes version 
 ✓ Kubernetes version "v1.25.2+4bd0702" is supported

 ✓ Checking Submariner support for the CNI network plugin
 ✓ The detected CNI network plugin ("OVNKubernetes") is supported
 ✓ Checking OVN version 
 ✓ The ovn-nb database version 6.3.0 is supported
 ✓ Checking gateway connections 
 ✓ All connections are established
 ✓ Globalnet deployment detected - checking if globalnet CIDRs overlap 
 ✓ Clusters do not have overlapping globalnet CIDRs
 ✓ Checking Submariner pods 
 ✓ All Submariner pods are up and running
⢀⡱ Checking if gateway metrics are accessible from non-gateway nodes  ⚠ Starting with Kubernetes 1.23, the Pod Security admission controller expects namespaces to have security labels. Without these, you will see warnings in subctl's output. subctl should work fine, but you can avoid the warnings and ensure correct behavior by adding these labels to the namespace submariner-operator:
 ⚠ pod-security.kubernetes.io/audit privileged
 ⚠ pod-security.kubernetes.io/warn privileged
 ✓ Checking if gateway metrics are accessible from non-gateway nodes 
 ✓ The gateway metrics are accessible
⠈⠑ Checking if globalnet metrics are accessible from non-gateway nodes  ⚠ Starting with Kubernetes 1.23, the Pod Security admission controller expects namespaces to have security labels. Without these, you will see warnings in subctl's output. subctl should work fine, but you can avoid the warnings and ensure correct behavior by adding these labels to the namespace submariner-operator:
 ⚠ pod-security.kubernetes.io/audit privileged
 ⚠ pod-security.kubernetes.io/warn privileged
 ✓ Checking if globalnet metrics are accessible from non-gateway nodes 
 ✓ The globalnet metrics are accessible
[...]

PR #127, opened 2022-06-24
Warn users about ns label warnings

Jaanki:
K8S1.23/OCP 4.11 and above throws warnings if the ns is not
`PodSecurity` labelled. This affects `diagnose kube-proxy-mode`
and `diagnose firewall` subcommands.

This PR adds a warning about these warnings so that the user
is not surprised when they pop up in the logs while running
these commands.

Partially Closes: #119

Depends on https://github.com/submariner-io/subctl/pull/126

Sample output when the `namespace` has not been specified by the user:
```
./cmd/bin/subctl diagnose firewall nat-discovery output/kubeconfigs/kind-config-cluster1 output/kubeconfigs/kind-config-cluster2
 ✓ Checking if nat-discovery port is opened on the gateway node of cluster "cluster1" 
 ✓ nat-discovery port is allowed in the cluster "cluster1"
```

Output when the namespace has been specified by the user:
```
./cmd/bin/subctl diagnose firewall nat-discovery output/kubeconfigs/kind-config-cluster1 output/kubeconfigs/kind-config-cluster2 --namespace user_ns
 ⚠ Kubernetes 1.23/OpenShift Container Platform 4.11 and above expects namespaces to have security labels as required by `PodSecurity Admission controller`. Otherwise, the related warnings would be logged. To avoid such warnings, make sure the namespace specified has below labels:
 ⚠ pod-security.kubernetes.io/enforce: privileged
 ⚠ pod-security.kubernetes.io/audit: privileged
 ⚠ pod-security.kubernetes.io/warn: privileged
```

Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

PR #349, opened 2022-11-02
subctl diagnose shows pod security warnings on OCP 4.12

tpantelis:
We set and check for all 3 of the `pod-security.kubernetes.io/*` namespace labels (`audit`, `warn`, and `enforce`) but, when set to the same mode, only one is really necessary since they have precedence. So it seems that OCP 4.12 removes the unnecessary `audit` and `warn` labels when `enforce` is set resulting in a warning from `subctl diagnose`. So the code was changed to only look for one of them and also to set only `enforce`.

From the K8s [Pod Security](https://kubernetes.io/docs/concepts/security/pod-security-admission/) docs:

Mode | Description
-- | --
**enforce** | Policy violations will cause the pod to be rejected.
**audit** | Policy violations will trigger the addition of an audit annotation to the event recorded in the audit log, but are otherwise allowed.
**warn** | Policy violations will trigger a user-facing warning, but are otherwise allowed.

Related to https://github.com/submariner-io/subctl/issues/348

### DIFF
--- a/internal/pods/schedule.go
+++ b/internal/pods/schedule.go
@@ -294,29 +294,29 @@ func checkNSLabels(config *Config) error {
 		"pod-security.kubernetes.io/warn":    "privileged",
 	}
 
-	missingLabels := map[string]string{}
+	foundOne := false
 
 	for key, expectedLabel := range expectedLabels {
 		actualLabel, found := ns.Labels[key]
-		if !found || actualLabel != expectedLabel {
-			missingLabels[key] = expectedLabel
+		if found && actualLabel == expectedLabel {
+			foundOne = true
+			break
 		}
 	}
 
-	if len(missingLabels) != 0 {
-		warnAbout(missingLabels, ns.Name)
+	if foundOne {
+		return nil
 	}
 
-	return nil
-}
+	labelStr := ""
+	for key, val := range expectedLabels {
+		labelStr += fmt.Sprintf("  %s=%s", key, val) + "\n"
+	}
 
-func warnAbout(missingLabels map[string]string, namespace string) {
 	status := cli.NewReporter()
 	status.Warning("Starting with Kubernetes 1.23, the Pod Security admission controller expects namespaces to have security labels."+
 		" Without these, you will see warnings in subctl's output. subctl should work fine, but you can avoid the warnings and ensure "+
-		"correct behavior by adding these labels to the namespace %s:", namespace)
+		"correct behavior by adding at least one of these labels to the namespace %q:\n%s", config.Namespace, labelStr)
 
-	for key, val := range missingLabels {
-		status.Warning(fmt.Sprintf("%s %s", key, val))
-	}
+	return nil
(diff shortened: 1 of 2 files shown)


skitt (approved):

nyechiel (approved):

internal/pods/schedule.go
```diff
@@ -294,29 +294,29 @@ func checkNSLabels(config *Config) error {
...
+	for key, val := range expectedLabels {
+		labelStr += fmt.Sprintf("  %s=%s", key, val) + "\n"
+	}
 
-func warnAbout(missingLabels map[string]string, namespace string) {
 	status := cli.NewReporter()
 	status.Warning("Starting with Kubernetes 1.23, the Pod Security admission controller expects namespaces to have security labels."+
 		" Without these, you will see warnings in subctl's output. subctl should work fine, but you can avoid the warnings and ensure "+
-		"correct behavior by adding these labels to the namespace %s:", namespace)
+		"correct behavior by adding at least one of these labels to the namespace %q:\n%s", config.Namespace, labelStr)
```
sridhargaddam, line 319: Does it mean that even if one of the label is present, its good enough?

tpantelis: I believe so in looking at the K8s docs but, then again, I'm not really sure of the history of this warning. We could just check for `enforce` I suppose. Or maybe the warning is no longer needed. @Jaanki did the original PR https://github.com/submariner-io/subctl/pull/127.

sridhargaddam (approved):

nyechiel:
I can confirm this is fixed with `0.14.0-rc2`:

```
$ subctl diagnose all
Cluster "maafried-cluster-a1"
 ✓ Checking Submariner support for the Kubernetes version 
 ✓ Kubernetes version "v1.25.2+4bd0702" is supported

 ✓ Checking Submariner support for the CNI network plugin
 ✓ The detected CNI network plugin ("OVNKubernetes") is supported
 ✓ Checking OVN version 
 ✓ The ovn-nb database version 6.3.0 is supported
 ✓ Checking gateway connections 
 ✓ All connections are established
 ✓ Non-Globalnet deployment detected - checking if cluster CIDRs overlap 
 ✓ Clusters do not have overlapping CIDRs
 ✓ Checking Submariner pods 
 ✓ All Submariner pods are up and running
 ✓ Checking if gateway metrics are accessible from non-gateway nodes 
 ✓ The gateway metrics are accessible
 ✓ Checking Submariner support for the kube-proxy mode 
 ✓ The kube-proxy mode is supported
 ✓ Checking the firewall configuration to determine if intra-cluster VXLAN traffic is allowed
 ✓ This check is not necessary for the OVNKubernetes CNI plugin
 ✓ The firewall configuration allows intra-cluster VXLAN traffic
 ✓ Globalnet is not installed - skipping

 ✓ Checking if services have been exported properly 
 ✓ All services have been exported properly
```

PR #403, opened 2022-11-24
Pull in the leases fix from the operator

skitt:


maayanf24 (approved):

sridhargaddam:
There are some failures in the lighthouse jobs, but its marked as not required. So approving the PR.

```
[Fail] [discovery] Test Stateful Sets Discovery Across Clusters when a pod tries to resolve a podname from stateful set in a remote cluster [It] should resolve the pod IP from the remote cluster 
github.com/submariner-io/shipyard@v0.13.1/test/e2e/framework/framework.go:517

[Fail] [discovery] Test Stateful Sets Discovery Across Clusters when a pod tries to resolve a podname from stateful set in a local cluster [It] should resolve the pod IP from the local cluster 
github.com/submariner-io/shipyard@v0.13.1/test/e2e/framework/framework.go:517

[Fail] [discovery] Test Stateful Sets Discovery Across Clusters when the number of active pods backing a stateful set changes [It] should only resolve the IPs from the active pods 
github.com/submariner-io/shipyard@v0.13.1/test/e2e/framework/framework.go:517
```

sridhargaddam (approved):

PR #427, opened 2022-12-07
Bump oauth2 to 0.3.0

skitt:
This dependency now has releases, bumping to a release will ensure dependabot tracks updates for us in future.


tpantelis (approved):

skitt:
dependabot bumped this to 0.20 _via_ `x/net`, so this is no longer useful (dependabot will presumably upgrade `oauth2` on its own).

PR #429, opened 2022-12-07
Improve diagnose firewall error output

tpantelis:
These commands spawn a pod to run a `tcpdump` command and reports a failure if the output doesn't contain an expected string. We should also report the pod output in case the `tcpdump` command fails.

For the `inter-cluster` command, the `tcpdump` output is piped to `grep` to match the expected client message and stop when found. However if an error occurs then the message isn't matched and therefore the pod output is empty which isn't useful. Use `sed` instead of `grep` which also outputs the non-matching lines.

### DIFF
--- a/pkg/diagnose/firewall.go
+++ b/pkg/diagnose/firewall.go
@@ -208,7 +208,7 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, names
 	// The following construct ensures that tcpdump will be stopped as soon as the message is seen, instead of waiting
 	// for a timeout; but when the message isn't seen, it will be killed once the timeout expires
 	podCommand := fmt.Sprintf(
-		"(tcpdump --immediate-mode -ln -Q in -A -s 100 -i any udp and %s & pid=\"$!\"; (sleep %d; kill \"$pid\") &) | grep -m1 '%s'",
+		"(tcpdump --immediate-mode -ln -Q in -A -s 100 -i any udp and %s & pid=\"$!\"; (sleep %d; kill \"$pid\") &) | sed '/%s/q'",
 		portFilter, options.ValidationTimeout, clientMessage)
 
 	sPod, err := spawnSnifferPodOnNode(localClusterInfo.ClientProducer.ForKubernetes(), gwNodeName, namespace, podCommand,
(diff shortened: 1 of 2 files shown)


pkg/diagnose/firewall.go
```diff
@@ -271,14 +271,13 @@ func verifyConnectivity(localClusterInfo, remoteClusterInfo *cluster.Info, names
...
-		status.Success("tcpdump output from sniffer pod on Gateway node")
-		status.Success(sPod.PodOutput)
+		status.Success("tcpdump output from sniffer pod on Gateway node:\n%s", sPod.PodOutput)
 	}
 
 	if !strings.Contains(sPod.PodOutput, clientMessage) {
 		return status.Error(fmt.Errorf("the tcpdump output from the sniffer pod does not include the message"+
 			" sent from client pod. Please check that your firewall configuration allows UDP/%d traffic"+
-			" on the %q node", destPort, localEndpoint.Spec.Hostname), "Error")
+			" on the %q node. Actual pod output: \n%s", destPort, localEndpoint.Spec.Hostname, sPod.PodOutput), "")
```
sridhargaddam: If you have a setup, can you paste the sample output in working/non-working scenario?

tpantelis: I don't currently have a setup and I didn't test with an actual setup where a firewall disallowed it but I did simulate it by expecting a bogus string.  Here's the output:

```
✗ The tcpdump output from the sniffer pod does not include the message sent from client pod. Please check that your firewall configuration allows UDP/4500 traffic on the "cluster1-worker" node. Actual pod output: 
  13:13:00.017573 eth0  In  IP 172.18.0.5.9898 > 172.18.0.7.4500: UDP-encap: ESP(spi=0x61383932,seq=0x61613837), length 1472
  E....; .@.;.........&... ..;a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa8
  13:13:00.017639 eth0  In  IP 172.18.0.5.9898 > 172.18.0.7.4500: UDP-encap: ESP(spi=0x39326161,seq=0x38370a61), length 808
  E..D.<@.@..<........&....0[r92aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
```

Do you think the output could be too much in a real env if it times out? If so I could truncate it at say 500 chars or so, ie a reasonable cushion to capture a failure like "Operation not permitted".

tpantelis: I truncated the output at 500. The above text is ~430 chars so wouldn't be truncated.

Here's an example of truncated output:

```
13:13:00.017573 eth0  In  IP 172.18.0.5.9898 > 172.18.0.7.4500: UDP-encap: ESP(spi=0x61383932,seq=0x61613837), length 1472
  E....; .@.;.........&... ..;a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa8
  13:13:00.017639 eth0  In  IP 172.18.0.5.9898 > 172.18.0.7.4500: UDP-encap: ESP(spi=0x39326161,seq=0x38370a61), length 808
  E..D.<@.@..<........&....0[r92aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
  a892aa87
  13:13:00.017639 eth0  In  IP 172.18.0.5.9898 > 172.18.0.7.4500: U... (truncated)
```

sridhargaddam: > If so I could truncate it at say 500 chars or so, ie a reasonable cushion to capture a failure like "Operation not permitted".

I'm just thinking if we really need 500 chars or can we reduce it to some lower value like 200 chars. I don't have the sample output when the "Operation not permitted" error is seen, but I guess even for such errors it should be within 200 chars.

sridhargaddam (approved):

pkg/diagnose/firewall.go
sridhargaddam: Anyways, lets go with 500 for now (as its logged only during errors). We can reduce it if necessary.

tpantelis: OK. The "Operation not permitted" error is 77 chars so well within either limit. I fiddled around with forcing other error scenarios which could yield much more output than that. We just need enough output so we could decipher the root cause - 500 seemed a pretty good cushion. The actual tcpdump output above truncated at 500 chars doesn't seem overwhelming to me.

aswinsuryan (approved):

PR #433, opened 2022-12-08
Give more permission to the diagnose pods

aswinsuryan:
Some containers which run os like rhel/fedora run tcpdump
as specific user id "72". So it needs pods to be privileged.
Also setting the runAsUser prevents the pods from starting with
random user id

* Make the diagnose pod a privileged
* Run the diagnose pod with userId 0

Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

### DIFF
--- a/internal/pods/schedule.go
+++ b/internal/pods/schedule.go
@@ -30,6 +30,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/pointer"
 )
 
 type schedulingType int
@@ -154,6 +155,12 @@ func (np *Scheduled) schedule() error {
 				Add:  []v1.Capability{"NET_ADMIN", "NET_RAW"},
 				Drop: []v1.Capability{"all"},
 			},
+			// Some containers which run os like rhel/fedora runs tcpdump
+			// as specific user id "72". So it needs pods to be privileged
+			// Also setting the runAsUser prevent the pods from starting with
+			// random user id
+			Privileged: pointer.Bool(true),
+			RunAsUser:  pointer.Int64(0),
 		}
 	}
 


tpantelis:
Can you explain in the summary why this is needed? Also perhaps a comment in the code.

tpantelis:
> Can you explain in the summary why this is needed? Also perhaps a comment in the code.

OK so this is to fix the `tcpdump` permission issue.

dfarrell07:
> Can you explain in the summary why this is needed? Also perhaps a comment in the code.

+1, let's try to be clear in the commit message and code about what fails exactly without this

aswinsuryan:
> > Can you explain in the summary why this is needed? Also perhaps a comment in the code.
> 
> +1, let's try to be clear in the commit message and code about what fails exactly without this

Done.

tpantelis (approved):

internal/pods/schedule.go
```diff
@@ -148,12 +148,21 @@ func (np *Scheduled) schedule() error {
 		},
 	}
 
+	// Some containers which run os like rhel/fedora runs tcpdump
+	// as specific user id "72". So it needs pods to be privileged.
+	// Also setting the runAsUser prevent the pods from starting with
+	// random user id
+	privileged := true
+	runAsUser := int64(0)
+
```
sridhargaddam: Okay, lets move this to Line159

Just for info, currently, we use privilege pods for two use-cases.
1. For identifying the hostname when it does not match with the name in the endpoint
2. As part of diagnose firewall validation 

In the other cases, when this API is called with podNetworking, we use commands like "curl" or "nc".

sridhargaddam (changes requested):

internal/pods/schedule.go
```diff
@@ -148,12 +148,21 @@ func (np *Scheduled) schedule() error {
...
 	if np.Config.Scheduling.Networking == HostNetworking {
 		networkPod.Spec.Containers[0].SecurityContext = &v1.SecurityContext{
 			Capabilities: &v1.Capabilities{
 				Add:  []v1.Capability{"NET_ADMIN", "NET_RAW"},
 				Drop: []v1.Capability{"all"},
 			},
+			Privileged: &privileged,
+			RunAsUser:  &runAsUser,
```
skitt: Please use `k8s.io/utils/pointer`:

```suggestion
			Privileged: pointer.Bool(true),
			RunAsUser:  pointer.Int64(0),
```

skitt (changes requested):

skitt (approved):

sridhargaddam (approved):

dfarrell07 (approved):

sridhargaddam:
@aswinsuryan I can handle the backport.

sridhargaddam:
Backport PR: https://github.com/submariner-io/subctl/pull/435

PR #434, opened 2022-12-09
Simplify getActiveGatewayNodeName

tpantelis:
This function enumerates the gateway nodes and spawns a pod to run the `hostname` command in search of the node whose name matches the local `Endpoint.Spec.Hostname`. However we can simplify this by, instead, searching for the `submariner-gateway` pod whose `gateway.submariner.io/status` label is set to "active" and obtaining its `gateway.submariner.io/node` label.

### DIFF
--- a/pkg/diagnose/firewall.go
+++ b/pkg/diagnose/firewall.go
@@ -30,6 +30,7 @@ import (
 	"github.com/submariner-io/subctl/pkg/cluster"
 	"github.com/submariner-io/subctl/pkg/image"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
+	"github.com/submariner-io/submariner-operator/pkg/names"
 	subv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/port"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pkg/diagnose/firewall.go
```diff
@@ -104,46 +105,21 @@ func spawnSnifferPodOnNode(client kubernetes.Interface, nodeName, namespace, pod
 	return spawnPod(client, scheduling, "validate-sniffer", namespace, podCommand, imageRepInfo)
 }
 
-func getActiveGatewayNodeName(clusterInfo *cluster.Info, hostname string, imageRepInfo *image.RepositoryInfo,
-	status reporter.Interface,
-) (string, error) {
-	nodes, err := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
-		LabelSelector: "submariner.io/gateway=true",
-	})
+func getActiveGatewayNodeName(clusterInfo *cluster.Info, status reporter.Interface) (string, error) {
```
sridhargaddam, line 108: This is a good optimisation @tpantelis. This would fail if the `submariner-gateway` pod is not running, but hey that's what diagnose is all for :)

sridhargaddam (approved):

skitt (approved):

aswinsuryan (approved):

PR #465, opened 2022-12-22
Ability to specify namespace for prefix context

Jaanki:
This patch adds an ability to specify a namespace for a prefixed context. Usage example:

`restconfig.NewProducer().WithPrefixedContext("broker").
    WithDefaultNamespace(constants.OperatorNamespace).
    WithPrefixedNamespace(constants.DefaultBrokerNamespace)`

Part of epic: https://github.com/submariner-io/enhancements/issues/143


internal/restconfig/restconfig.go
```diff
@@ -72,6 +72,8 @@ type Producer struct {
 	deprecatedKubeContextsMessage *string
 	defaultNamespace              *string
 	deprecatedNamespaceFlag       *string
+	prefixedNamespace             *string
```
skitt: Please make it clear that this is the default namespace. Since multiple prefixes can be defined, this will have to be a map (see `prefixedKubeConfigs`), _e.g._ `prefixedDefaultNamespaces`.

internal/restconfig/restconfig.go
```diff
@@ -72,6 +72,8 @@ type Producer struct {
...
+	prefixedNamespaceFlag         bool
```
skitt: This isn’t necessary: enabling the namespace flag automatically adds prefixed namespace flags.

skitt (changes requested):

internal/restconfig/restconfig.go
Jaanki: what if `namespace` flag is not enabled? i. e the default ns is not needed. Just the prefixed ns.

skitt: If we ever need that we can add it then.

internal/restconfig/restconfig.go
```diff
@@ -107,6 +108,15 @@ func (rcp *Producer) WithDefaultNamespace(defaultNamespace string) *Producer {
 	return rcp
 }
 
+// WithPrefixedNamespace configures the producer to set up a prefixed namespace flag,
+// with the given value.
```
skitt:
```suggestion
// with the given default value.
```

internal/restconfig/restconfig.go
```diff
@@ -335,8 +345,8 @@ func (rcp *Producer) RunOnSelectedPrefixedContext(prefix string, function PerCon
 			return true, status.Error(err, "error retrieving the namespace for the configuration for prefix %s", prefix)
 		}
 
-		if !overridden && rcp.defaultNamespace != nil {
-			namespace = *rcp.defaultNamespace
+		if !overridden && rcp.prefixedDefaultNamespaces[prefix] != nil {
+			namespace = *rcp.prefixedDefaultNamespaces[prefix]
 		}
```
skitt: This should still allow falling back to the default “global” namespace (`subctl diagnose` expects this):

```suggestion
		if !overridden {
			if rcp.prefixedDefaultNamespaces[prefix] != nil {
				namespace = *rcp.prefixedDefaultNamespaces[prefix]
			} else if rcp.defaultNamespace != nil {
				namespace = *rcp.defaultNamespace
			}
		}
```

skitt (changes requested):

skitt (approved):

tpantelis (approved):

PR #466, opened 2022-12-23
Switch to ginkgo v2

skitt:
Depends on https://github.com/submariner-io/shipyard/pull/1105
Depends on https://github.com/submariner-io/cloud-prepare/pull/533
Depends on https://github.com/submariner-io/lighthouse/pull/1021
Depends on https://github.com/submariner-io/submariner-operator/pull/2431
Depends on https://github.com/submariner-io/shipyard/pull/1105


dfarrell07 (approved):

tpantelis (approved):

tpantelis:
@skitt We need a change in **shipyard** `e2e.RunE2ETests` to pass in the **ginkgo** `SuiteConfig` and `ReporterConfig`. I can do that tomorrow.

tpantelis:
@skitt FYI: the `SlowSpecThreshold` field is no longer available in the `ReporterConfig`. The only way to set it is via CLI but it's been deprecated altogether:  "_--slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck._"

skitt:
Yup, that’s why I dropped it in Shipyard.

PR #474, opened 2023-01-02
Remove all deprecated commands and flags

skitt:
This adjusts the system tests to ensure the old forms produce an
error. subctl diagnose errors are ignored on non-deprecated forms;
this doesn't change the test coverage compared to previous versions,
because the _subctl function used to ignore errors entirely.


tpantelis (approved):

scripts/test/system.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -em -o pipefail
```
mkolesnik: I don't think it's a good idea to drop these, if a command now fails it wont fail the test and we wont know that something broke.

Instead you could define a function that calls `_subctl` and expects it to fail.

For example, I broke the tests but they still passed:
```bash
::group::Running 'subctl uninstall -y --context cluster2 --fdsafs'
[06:32:07.131] [dir=subctl; cl=cluster2; fn=_subctl]$ /go/src/github.com/submariner-io/subctl/cmd/bin/subctl uninstall -y --context cluster2 --fdsafs
Error: unknown flag: --fdsafs
Usage:
  subctl uninstall [flags]

Flags:
      --as string                      Username to impersonate for the operation
...
  -y, --yes                            automatically answer yes to confirmation prompt

unknown flag: --fdsafs
[06:32:07.151] [dir=subctl; cl=cluster2; fn=_subctl]$ ret=1
::endgroup::
[06:32:07.152] [dir=subctl; cl=cluster2; fn=_subctl]$ return 1
...
Successfully remade target file 'system-test'.
[mkolesni@mkolaptop subctl ((69276fc...) *)]$ echo $?
0
```

cmd/subctl/diagnose.go
```diff
@@ -237,37 +233,34 @@ func diagnoseAll(status reporter.Interface) error {
...
-			}, status))
+			}
+			if !found {
+				return errors.New("no remote context was specified")
+			}
+			return nil
+		}, status), "Error running command")
+}
+
+func checkNoArguments(cmd *cobra.Command, args []string) error {
```
mkolesnik: nit: This isn't specific to `diagnose`, a better place would be in `root.go`

mkolesnik (changes requested):

scripts/test/system.sh
skitt: The problem is that `_subctl` ignored errors previously anyway. Fixing `_subctl` so that it no longer ignores errors, which I need so that the tests can actually verify that certain forms of `subctl` fail, causes the overall system test to fail earlier because `subctl` exits with an error in certain “normal” cases anyway.

Since `set -e -o pipefail` was effectively meaningless, I removed it; we can add it back later, once we have a consistent set of rules for `subctl`’s exit code.

dfarrell07 (approved):

scripts/test/system.sh
mkolesnik: I don't understand what you mean by
> `_subctl` ignored errors previously anyway

and
> Since `set -e -o pipefail` was effectively meaningless

It didn't ignore errors, it would fail the tests due to any error, eg try to apply this and run the system tests:

```patch
diff --git a/scripts/test/system.sh b/scripts/test/system.sh
index 69a4e9c..62d470d 100755
--- a/scripts/test/system.sh
+++ b/scripts/test/system.sh
@@ -133,7 +133,7 @@ deploy_env_once
 
 # Test subctl show invocations
 
-_subctl show all | tee /dev/stderr | (sponge ||:) | grep -q 'Cluster "cluster2"'
+_subctl show all --sdfsdf | tee /dev/stderr | (sponge ||:) | grep -q 'Cluster "cluster2"'
 # Single-context variants don't say 'Cluster "foo"', check what cluster is considered local
 _subctl show all --context cluster1 | tee /dev/stderr | (sponge ||:) | grep -qv 'cluster2.*local'
 _subctl show all --context cluster2 | tee /dev/stderr | (sponge ||:) | grep -q 'cluster2.*local'
```

This is just to force a failure, or are you claiming that `subctl` itself doesn't produce any error codes?

skitt: Ah, I was barking up the wrong tree, it’s not that `_subctl` ignores errors, it’s the combination of functions and `&&` again :-/.

mkolesnik: Yes that one is indeed problematic but it doesn't have anything to do with the `set -e`..

You can make this change to ensure it (as we don't care about the specific exit code):
```suggestion
    "${DAPPER_SOURCE}"/cmd/bin/subctl "$@" || exit 1
```

GHA will be fine without a closing `::endgroup::` anyway, and the code will be tidier :)

scripts/test/system.sh
```diff
@@ -22,10 +22,13 @@ function deploy_env_once() {
 }
 
 function _subctl() {
+    local ret
     # Print GHA groups to make looking at CI output easier
     echo "::group::Running 'subctl $*'"
     "${DAPPER_SOURCE}"/cmd/bin/subctl "$@"
+    ret=$?
     echo "::endgroup::"
+    return $ret
```
mkolesnik: If you happen to respin this, otherwise you can leave as is.

You can make this change to ensure it exits (GHA will be fine without a closing `::endgroup::` anyway) and the code will be tidier :)

```suggestion
    # Print GHA groups to make looking at CI output easier
    echo "::group::Running 'subctl $*'"
    "${DAPPER_SOURCE}"/cmd/bin/subctl "$@" || exit "$?"
    echo "::endgroup::"
```

mkolesnik (approved):

Jaanki (approved):

scripts/test/system.sh
```diff
@@ -24,7 +24,7 @@ function deploy_env_once() {
 function _subctl() {
     # Print GHA groups to make looking at CI output easier
     echo "::group::Running 'subctl $*'"
-    "${DAPPER_SOURCE}"/cmd/bin/subctl "$@"
+    "${DAPPER_SOURCE}"/cmd/bin/subctl "$@" || exit "$?"
```
mkolesnik:
```suggestion
    "${DAPPER_SOURCE}"/cmd/bin/subctl "$@" || return "$?"
```

Sorry should've been `return` as it isn't running in a sub-shell

PR #491, opened 2023-01-10
Initialize `prefixedDefaultNamespaces` before using

Jaanki:
otherwise Go throws runtime error:
`panic: assignment to entry in nil map`

EPIC: https://github.com/submariner-io/enhancements/issues/143


internal/restconfig/restconfig.go
```diff
@@ -112,6 +112,7 @@ func (rcp *Producer) WithDefaultNamespace(defaultNamespace string) *Producer {
 // with the given default value.
 // The chosen namespace will be passed to the PerContextFn used to process the context.
 func (rcp *Producer) WithPrefixedNamespace(prefix, prefixedNamespace string) *Producer {
+	rcp.prefixedDefaultNamespaces = make(map[string]*string)
```
skitt: This should be in `NewProducer`, otherwise every call to `WithPrefixedNamespace` will clear any existing prefixed default namespace.

skitt (changes requested):

internal/restconfig/restconfig.go
Jaanki: or `SetupFlags`?

skitt: That’s called after `WithPrefixedNamespace`, so it wouldn’t avoid the panic (and it would lose any prefixed default namespace as well).

internal/restconfig/restconfig.go
```diff
@@ -75,7 +75,7 @@ type Producer struct {
 
 // NewProducer initialises a blank producer which needs to be set up with flags (see SetupFlags).
 func NewProducer() *Producer {
-	return &Producer{}
+	return &Producer{prefixedDefaultNamespaces: map[string]*string{}}
```
skitt:
```suggestion
	return &Producer{prefixedDefaultNamespaces: make(map[string]*string)}
```

internal/restconfig/restconfig.go
skitt:
```suggestion
```

skitt (changes requested):

internal/restconfig/restconfig.go
skitt: You still need to delete this.

skitt (changes requested):

skitt (approved):

tpantelis (approved):

PR #493, opened 2023-01-11
Recover broker-info.subm

Jaanki:
from the specified Broker and Submariner clusters.

IPSec PSK secret is first looked for in a Submariner CR, then from the file user provided and then generated new.

Epic: https://github.com/submariner-io/enhancements/issues/143

### DIFF
--- a/pkg/broker/reconstruct.go
+++ b/pkg/broker/reconstruct.go
@@ -1,0 +1,72 @@
+package broker
+
+import (
+	"context"
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/reporter"
+	"github.com/submariner-io/subctl/internal/constants"
+	"github.com/submariner-io/subctl/internal/rbac"
+	"github.com/submariner-io/subctl/pkg/cluster"
+	"k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ReconstructData(
+	brokerCluster, submCluster *cluster.Info, namespace string, status reporter.Interface,
+	ipsecSubmFile string) error {
+	status.Start("Retrieving data to reconstruct broker-info.subm")
+    defer status.End()
+
+	data := &Info{}
+	var err error
+
+	data.BrokerURL = brokerCluster.RestConfig.Host + brokerCluster.RestConfig.APIPath
+
+	data.ClientToken, err = rbac.GetClientTokenSecret(
+		context.TODO(), brokerCluster.ClientProducer.ForKubernetes(), namespace,
+		constants.SubmarinerBrokerAdminSA,
+	)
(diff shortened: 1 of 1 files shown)


pkg/broker/reconstruct.go
```diff
@@ -0,0 +1,83 @@
...
+	brokerClusterInfo, submClusterInfo *cluster.Info, namespace string, status reporter.Interface,
+	ipsecSubmFile string) error {
+	status.Start("Retrieving data to reconstruct broker-info.subm")
+    defer status.End()
+
+	data := &Info{}
+
+	kubeClient, err := kubernetes.NewForConfig(brokerClusterInfo.RestConfig)
+	if err != nil {
+		return status.Error(err, "error creating Kubernetes client")
```
skitt:
```suggestion
		return status.Error(err, "error creating Kubernetes client to access the broker")
```

pkg/broker/reconstruct.go
```diff
@@ -0,0 +1,83 @@
...
+	} else if ipsecSubmFile != "" {
+		status.Success("Retrieving IPSec PSK from the file %s", ipsecSubmFile)
+		ipsecData, err := ReadInfoFromFile(ipsecSubmFile)
+		if err != nil {
+			return errors.Wrapf(err, "error importing IPsec PSK secret from file %q", ipsecSubmFile)
+		}
+
+		data.IPSecPSK = ipsecData.IPSecPSK
+	} else {
+		status.Success("Generating new IPSec PSK secret")
```
skitt, line 60: This should be a warning, because the new file won’t be usable to connect existing clusters.

skitt (changes requested):

pkg/broker/reconstruct.go
```diff
@@ -0,0 +1,83 @@
...
+		Name: brokercr.Name,
+	}, broker)
+
+	if err == nil {
+		data.Components = broker.Spec.Components
+		data.ServiceDiscovery = data.IsServiceDiscoveryEnabled()
+		data.CustomDomains = &broker.Spec.DefaultCustomDomains
+	} else if !apierrors.IsNotFound(err) && !meta.IsNoMatchError(err) {
+		return status.Error(err, "error retrieving Broker")
+	}
```
skitt: If we can’t retrieve the `Broker` object, that’s a fatal error in all cases:

```suggestion
	if err != nil {
		return status.Error(err, "error retrieving Broker")
	}

	data.Components = broker.Spec.Components
	data.ServiceDiscovery = data.IsServiceDiscoveryEnabled()
	data.CustomDomains = &broker.Spec.DefaultCustomDomains
```

pkg/broker/reconstruct.go
Jaanki: This scenario will come only when there are no existing joined clusters.

pkg/broker/reconstruct.go
Jaanki: This code will be executed only when `broker` is present on the cluster.

Jaanki:
Closing this in favor of https://github.com/submariner-io/subctl/pull/467

PR #518, opened 2023-01-26
Only call cloud prepare when it's needed

### LINKED ISSUE #517: Failed to prepare GCP cloud
**What happened**:

```
$ subctl cloud prepare gcp --ocp-metadata cluster-a/metadata.json
 ✓ Obtained infra ID "nyechiel-cluster-a-tbc8x", region "us-east1", and project ID "mw-integration-submariner" from OCP metadata file "cluster-a/metadata.json"
 ✓ Retrieving GCP credentials from your GCP configuration
 ✓ Initializing GCP connectivity
 ✓ Configuring the required firewall rules for inter-cluster traffic 
 ✓ Opened External ports "4500/udp, 4490/udp, 0/esp, 0/ah" with firewall rule "nyechiel-cluster-a-tbc8x-submariner-public-ports-ingress" on GCP
 ✓ Retrieving the current zones in the project
 ✓ Retrieved the zones
 ✓ Verifying if current gateways match the required number of gateways 
 ✓ Current gateways match the required number of gateways
 ✗ Opening internal ports "" for intra-cluster communications on GCP 
 ✗ Unable to open ports: error inserting firewall rule &compute.Firewall{Allowed:[]*compute.FirewallAllowed{}, CreationTimestamp:"", Denied:[]*compute.FirewallDenied(nil), Description:"", DestinationRanges:[]string(nil), Direction:"INGRESS", Disabled:false, Id:0x0, Kind:"", LogConfig:(*compute.FirewallLogConfig)(nil), Name:"nyechiel-cluster-a-tbc8x-submariner-internal-ports-ingress", Network:"projects/mw-integration-submariner/global/networks/nyechiel-cluster-a-tbc8x-network", Priority:0, SelfLink:"", SourceRanges:[]string(nil), SourceServiceAccounts:[]string(nil), SourceTags:[]string{"nyechiel-cluster-a-tbc8x-worker", "nyechiel-cluster-a-tbc8x-master"}, TargetServiceAccounts:[]string(nil), TargetTags:[]string{"nyechiel-cluster-a-tbc8x-worker", "nyechiel-cluster-a-tbc8x-master"}, ServerResponse:googleapi.ServerResponse{HTTPStatusCode:0, Header:http.Header(nil)}, ForceSendFields:[]string(nil), NullFields:[]string(nil)}: googleapi: Error 400: Invalid value for field 'resource.allowed[0]': ''. Must specify at least one rule., invalid
 ✗ Failed to prepare GCP cloud: unable to open ports: error inserting firewall rule &compute.Firewall{Allowed:[]*compute.FirewallAllowed{}, CreationTimestamp:"", Denied:[]*compute.FirewallDenied(nil), Description:"", DestinationRanges:[]string(nil), Direction:"INGRESS", Disabled:false, Id:0x0, Kind:"", LogConfig:(*compute.FirewallLogConfig)(nil), Name:"nyechiel-cluster-a-tbc8x-submariner-internal-ports-ingress", Network:"projects/mw-integration-submariner/global/networks/nyechiel-cluster-a-tbc8x-network", Priority:0, SelfLink:"", SourceRanges:[]string(nil), SourceServiceAccounts:[]string(nil), SourceTags:[]string{"nyechiel-cluster-a-tbc8x-worker", "nyechiel-cluster-a-tbc8x-master"}, TargetServiceAccounts:[]string(nil), TargetTags:[]string{"nyechiel-cluster-a-tbc8x-worker", "nyechiel-cluster-a-tbc8x-master"}, ServerResponse:googleapi.ServerResponse{HTTPStatusCode:0, Header:http.Header(nil)}, ForceSendFields:[]string(nil), NullFields:[]string(nil)}: googleapi: Error 400: Invalid value for field 'resource.allowed[0]': ''. Must specify at least one rule., invalid
```

All preparation steps to be completed with no errors.

1. Deploy OCP on top of GCP. I am using OpenShift 4.12.0 (Kubernetes Version: v1.25.4+77bec7a)
2. Export kubefonfig
3. Install `subctl`. I am using 0.15.0-m2
4. Run the `subctl cloud prepare gcp` command

mkolesnik:
Since the cloud prepare step opens internal ports and nothing else, we should only call it when there are ports to open.

Resolves #517 

### DIFF
--- a/pkg/cloud/prepare/aws.go
+++ b/pkg/cloud/prepare/aws.go
@@ -54,7 +54,11 @@ func AWS(clusterInfo *cluster.Info, ports *cloud.Ports, config *aws.Config, useL
 				}
 			}
 
-			return cloud.PrepareForSubmariner(input, status)
+			if len(input.InternalPorts) > 0 {
+				return cloud.PrepareForSubmariner(input, status)
+			}
+
+			return nil
 		})
 
 	return status.Error(err, "Failed to prepare AWS cloud")
--- a/pkg/cloud/prepare/azure.go
+++ b/pkg/cloud/prepare/azure.go
@@ -51,7 +51,11 @@ func Azure(clusterInfo *cluster.Info, ports *cloud.Ports, config *azure.Config, 
 				}
 			}
 
-			return cloud.PrepareForSubmariner(input, status)
+			if len(input.InternalPorts) > 0 {
+				return cloud.PrepareForSubmariner(input, status)
+			}
+
+			return nil
 		})
 
 	return status.Error(err, "Failed to prepare Azure  cloud")
--- a/pkg/cloud/prepare/gcp.go
+++ b/pkg/cloud/prepare/gcp.go
@@ -48,7 +48,11 @@ func GCP(clusterInfo *cluster.Info, ports *cloud.Ports, config *gcp.Config, useL
 				}
 			}
 
-			return cloud.PrepareForSubmariner(input, status)
+			if len(input.InternalPorts) > 0 {
+				return cloud.PrepareForSubmariner(input, status)
+			}
+
+			return nil
 		})
 
 	return status.Error(err, "Failed to prepare GCP cloud")
--- a/pkg/cloud/prepare/rhos.go
+++ b/pkg/cloud/prepare/rhos.go
@@ -49,7 +49,11 @@ func RHOS(clusterInfo *cluster.Info, ports *cloud.Ports, config *rhos.Config, us
 				}
 			}
 
-			return cloud.PrepareForSubmariner(input, status)
+			if len(input.InternalPorts) > 0 {
+				return cloud.PrepareForSubmariner(input, status)
+			}
+
+			return nil
 		})
 
 	return status.Error(err, "Failed to prepare RHOS cloud")


pkg/cloud/prepare/aws.go
```diff
@@ -54,7 +54,11 @@ func AWS(clusterInfo *cluster.Info, ports *cloud.Ports, config *aws.Config, useL
 				}
 			}
 
-			return cloud.PrepareForSubmariner(input, status)
+			if len(input.InternalPorts) > 0 {
+				return cloud.PrepareForSubmariner(input, status)
+			}
```
tpantelis, line 59: It seems like this validation should be done in `cloud.PrepareForSubmariner` otherwise this code is making assumptions about what `cloud.PrepareForSubmariner` does internally, ie calling it in a 'white box' manner rather than 'black box'.  We'd have a latent bug if the behavior of `cloud.PrepareForSubmariner` ever changed such that it did more than open internal ports.

tpantelis (changes requested):

pkg/cloud/prepare/aws.go
mkolesnik: Actually right now (and for a long while) it's only opening internal ports so it makes sense to do so here and rename that function in cloud prepare itself if we're worried about that.

skitt: We know that, but `subctl` isn’t supposed to. The description of `PrepareForSubmariner` is very generic, and just says it prepares a cloud for Submariner. The AWS version already handles this correctly, so there is precedent for handling empty `InternalPorts` slices in the `PrepareForSubmariner` function — and I’d argue that *that* makes sense, there’s no documented requirement that `PrepareForSubmariner` be called only with non-empty `InternalPorts`.

mkolesnik: I propose we go with https://github.com/submariner-io/cloud-prepare/pull/571 and leave this as is.

In the same manner, we only try to deploy gateways when gateways have been requested.

The whole `PrepareForSubmariner` made sense initially when cloud prepare was just incepted, but due to limitations it has been split out to internal & external preparation, and realistically ever since it started it just opens internal ports.
Using a simpler API makes it clear and avoids any confusion that might arise, while keeping it simple.

skitt (approved):

pkg/cloud/prepare/aws.go
tpantelis: It's fine to rename the API but I still think the implementation should handle the trivial case of an empty port list to avoid all users having to do it. As @skitt mentioned, the AWS implementation does check but others don't - we should be consistent.

nyechiel:
Do we also need this backported, and if so, to what branches?

skitt:
> Do we also need this backported, and if so, to what branches?

I vote for backporting it to all supported branches, 0.12, 0.13, and 0.14.

aswinsuryan (approved):

pkg/cloud/prepare/aws.go
aswinsuryan: I think while it needs to be fixed in respective implementations as well, we can have this check too

tpantelis (approved):

skitt:
Manually backported to 0.12 as https://github.com/submariner-io/submariner-operator/pull/2473.

PR #542, opened 2023-02-02
Force delete objects on uninstall if operator pod not running

### LINKED ISSUE #31: `subctl uninstall` times out if the operator isn't running
**What happened**:

`subctl uninstall` timed out while deleting the `ServiceDiscovery` resource:

 ```
✗ Deleting the ServiceDiscovery resource - this may take some time
 ✗ Error deleting ServiceDiscovery resource "service-discovery": timed out waiting for the condition
```

This can happen if the operator isn't running. Ideally it should check that the operator deployment is installed and running. If not, then log a warning and force delete the resource by removing the finalizer.

Delete the operator deployment before running `subctl uninstall`

### PR DESCRIPTION
**Pull request by tpantelis**
The operator is supposed to perform cleanup and remove the finalizer on deletion of the `Submariner` and `ServiceDiscovery` resources. But if it's not running, uninstall would timeout and fail. So uninstall was modified to check if the operator pod is running and, if not, log a warning and force delete the resource by removing the finalizer.

Fixes https://github.com/submariner-io/subctl/issues/31

Depends on https://github.com/submariner-io/submariner-operator/pull/2485
Depends on https://github.com/submariner-io/subctl/pull/544

### DIFF
--- a/pkg/uninstall/all.go
+++ b/pkg/uninstall/all.go
@@ -20,20 +20,25 @@ package uninstall
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/finalizer"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/util"
 	"github.com/submariner-io/subctl/internal/constants"
 	"github.com/submariner-io/subctl/pkg/client"
+	"github.com/submariner-io/subctl/pkg/operator/deployment"
 	operatorv1alpha1 "github.com/submariner-io/submariner-operator/api/v1alpha1"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/uninstall/all.go [hunk 1]
```diff
@@ -237,24 +241,62 @@ func ensureServiceDiscoveryDeleted(controllerClient controller.Client, clusterNa
...
 		}
 
-		return false, err
-	})
+		podStatusStr := ""
+		if len(pods.Items) == 0 {
+			podStatusStr = "does not exist"
+		} else {
+			if pods.Items[0].Status.Phase == corev1.PodRunning {
+				return status.Error(fmt.Errorf("the Submariner operator pod appears to be running but did not"+
```
skitt:
```suggestion
				return status.Error(fmt.Errorf("the Submariner operator pod appears to be running but did not "+
```

skitt (changes requested):
Just a missing space.

skitt (approved):

skitt (approved):

dfarrell07 (approved):

PR #544, opened 2023-02-03
Obtain label selector for gathering operator logs

### LINKED ISSUE #490: subctl gather doesn’t gather operator logs
`subctl gather` should gather the operator’s logs, but it doesn’t.

tpantelis:
If the operator `Deployment` is not created via `subctl join`, we can't assume it has the `name=submariner-operator` label so obtain the label to select from the `Deployment` resource.

Fixes https://github.com/submariner-io/subctl/issues/490


skitt (approved):

dfarrell07 (approved):

dfarrell07:
This went in before release-0.15 was cut, so it doesn't need to be backported there before RC0.

PR #551, opened 2023-02-07
Add support for packet size in subctl verify

### LINKED ISSUE #499: Enhance subctl verify to validate mtu between the clusters
As part of `subctl verify ...` command we execute various datapath tests to validate the connectivity between the clusters.
Sometimes the connections could be successful but we might have MTU issues. 

From the output of `subctl verify ...`, to some extent we are able to figure out if its MTU issue or connection failure by looking into the verbose logs, but its not easy to figure it out and also we may not have verbose logs all the time.

Having a dedicated sub-command to validate the MTU will help us to easily identify a datapath failure is because of connection failures or if its due to MTU issues between the clusters.

yboaron:
Allow specifying packet size in e2e tcp connectivity tests,
this may help with troubleshooting MTU issues.

Depends on: https://github.com/submariner-io/shipyard/pull/1140
Fixes: https://github.com/submariner-io/subctl/issues/499

Signed-off-by: Yossi Boaron <yboaron@redhat.com>


cmd/subctl/verify.go
```diff
@@ -142,6 +143,7 @@ func addVerifyFlags(cmd *cobra.Command) {
 		"namespace in which submariner is deployed")
 	cmd.Flags().StringVar(&verifyOnly, "only", strings.Join(getAllVerifyKeys(), ","), "comma separated verifications to be performed")
 	cmd.Flags().BoolVar(&disruptiveTests, "disruptive-tests", false, "enable disruptive verifications like gateway-failover")
+	cmd.Flags().BoolVar(&useSmallPackets, "use-small-packets", false, "use small packets in connectivity verification")
```
sridhargaddam: instead of use-small-packets, what are your thoughts on supporting packet-size?

tpantelis: 👍 That was my initial thought as well...

yboaron: I assume that supporting packet-size means we should update shipyard code ( check [1] and [2] ) as currently, we send data-buffer (sized  16bytes uuid converted to string 'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx' ) multiple X  times. 

I personally don't think we should be that flexible, I assume 2 packet size , normal and small should do the work for us. 

Any special use cases that you think flexible packet size can help with? 
Let me know If you think supporting normal/small packet size isn't sufficient  


[1] 
https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/network_pods.go#L106
[2]
https://github.com/submariner-io/shipyard/blob/devel/test/e2e/framework/network_pods.go#L269

tpantelis: I guess if there's a case where hard-coded normal/small packet sizes don't suffice?

yboaron: OK, I changed the flag to packet size

cmd/subctl/verify.go
```diff
@@ -142,6 +143,8 @@ func addVerifyFlags(cmd *cobra.Command) {
 		"namespace in which submariner is deployed")
 	cmd.Flags().StringVar(&verifyOnly, "only", strings.Join(getAllVerifyKeys(), ","), "comma separated verifications to be performed")
 	cmd.Flags().BoolVar(&disruptiveTests, "disruptive-tests", false, "enable disruptive verifications like gateway-failover")
+	cmd.Flags().UintVar(&packetSize, "packet-size", 3000, "size of packet")
```
sridhargaddam: Is the default packet-size of 3000 matching the current implementation?

yboaron: Correct, with size equals to 3000 we'll echo 50 times [1]  data buffer which is similar to the current status.


[1] 
https://github.com/submariner-io/shipyard/pull/1140/files#diff-8e8c299993caac0264030c88915843d710ec8f4ef455082beb1683e2b9cbc27cR276

yboaron: BTW, any idea why we use 3000 bytes for TCP payload ?  I guess that any value > 1500 will cover MTU issues in networks with MTU=1500, and it should shorten the test time

sridhargaddam: AFAICT there is no specific reason behind 3000 bytes. Reducing the payload may not really help in shortening the test time when the connections are good. The time difference between 3000 bytes vs 1600 bytes may not be significant, the real difference will be when there is connection failure and the values of `--connection-attempts` and `--connection-timeout` come into picture. Sometimes, while testing, I use custom values for these parameters to shorten the time interval.

skitt (approved):

cmd/subctl/verify.go
tpantelis: Can we provide more help description than just "size of packet" ?

nyechiel (approved):

PR #553, opened 2023-02-08
Print clusterIDs instead of cluster name from kubeconfig

Jaanki:
whenever possible. So when a cluster is joined with the user-specified clusterID (via --clusterid flag), all subctl commands will print that clusterID and not the clustername from kubeconfig.


dfarrell07:
Do we want to un-draft this and discuss?

skitt:
This needs to be addressed more broadly. We need to separate context names (defined by the user in their kubeconfig) from cluster ids (used in the broker), and ensure that all references to either of these use the correct value.

PR #570, opened 2023-02-23
Failed to diagnose calico when natOutgoing is not found

cyclinder:
```
[root@master1 subctl-devel]# subctl-devel diagnose cni
Cluster "cluster.local"
 ✓ Checking Submariner support for the CNI network plugin
 ✓ The detected CNI network plugin ("calico") is supported
 ✗ Calico CNI detected, checking if the Submariner IPPool pre-requisites are configured
 ✗ natOutgoing status not found for IPPool "cluster2-globalcidr"

subctl version: devel

[root@master1 subctl-devel]# calicoctl get ippools -o wide
NAME                  CIDR            NAT     IPIPMODE   VXLANMODE   DISABLED   DISABLEBGPEXPORT   SELECTOR
cluster2-globalcidr   242.1.10.0/24   false   Never      Never       true       false              all()
default-pool          172.20.0.0/16   true    Never      Always      false      false              all()
default-pool-ipv6     fc01::/112      false   Never      Always      false      false              all()

[root@master1 subctl-devel]# calicoctl get ippools cluster2-globalcidr -o yaml
apiVersion: projectcalico.org/v3
kind: IPPool
metadata:
  creationTimestamp: "2023-02-22T14:56:58Z"
  name: cluster2-globalcidr
  resourceVersion: "42048"
  uid: 3462d93f-ba53-4971-b550-083fa0b65dfb
spec:
  allowedUses:
  - Workload
  - Tunnel
  blockSize: 26
  cidr: 242.1.10.0/24
  disabled: true
  ipipMode: Never
  nodeSelector: all()
  vxlanMode: Never
```

`natOutgoing` is of bool type so it should be considered false when not found.

### DIFF
--- a/pkg/diagnose/cni.go
+++ b/pkg/diagnose/cni.go
@@ -228,7 +228,8 @@ func getSpecBool(pool unstructured.Unstructured, key string) (bool, error) {
 	}
 
 	if !found {
-		return false, fmt.Errorf("%s status not found for IPPool %q", key, pool.GetName())
+		// for bool type, not found is considered false
+		return false, nil
 	}
 
 	return isDisabled, nil


pkg/diagnose/cni.go
```diff
@@ -228,7 +228,8 @@ func getSpecBool(pool unstructured.Unstructured, key string) (bool, error) {
 	}
 
 	if !found {
-		return false, fmt.Errorf("%s status not found for IPPool %q", key, pool.GetName())
+		// for bool type, not found is considered isDisabled
+		return true, nil
```
tpantelis: This should be false.

```suggestion
		// for bool type, not found is considered false
		return false, nil
```

tpantelis (changes requested):

pkg/diagnose/cni.go
cyclinder: Thanks @tpantelis for the review. Updated.

tpantelis (approved):

dfarrell07 (approved):

skitt (approved):

dfarrell07:
Release notes here: https://github.com/submariner-io/submariner-website/pull/928

PR #575, opened 2023-02-27
Deprecate `dedicated-gateway` for 0.15

### LINKED ISSUE #603: Deprecate non-dedicated gateways on 0.15
**What would you like to be removed**:
Let's deprecate this mode for 0.15 and remove it on 0.16

**Why is this needed**:
We're not actually testing or using non-dedicated gateway mode from cloud prepare.
Furthermore, a more K8s native approach is to deploy with LB mode which actually handles the cloud related operations properly.

From a technical debt perspective, the code is just sitting there and making cloud-prepare and subctl harder to maintain, with it removed we'll have an easier maintenance burden and could even further simplify the cloud prepare code.

mkolesnik:
We're not actually testing or using non-dedicated gateway mode from cloud prepare.
Furthermore, a more K8s native approach is to deploy with LB mode which actually handles the cloud related operations properly.

From a technical debt perspective, the code is just sitting there and making cloud-prepare and subctl harder to maintain, with it removed we'll have an easier maintenance burden and could even further simplify the cloud prepare code.

Resolves [#603](https://github.com/submariner-io/cloud-prepare/issues/603)

### DIFF
--- a/cmd/subctl/azure.go
+++ b/cmd/subctl/azure.go
@@ -79,6 +79,9 @@ func init() {
 		"Whether a dedicated gateway node has to be deployed")
 	cloudPrepareCmd.AddCommand(azurePrepareCmd)
 
+	_ = azurePrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
+
 	addGeneralAzureFlags(azureCleanupCmd)
 	cloudCleanupCmd.AddCommand(azureCleanupCmd)
 }
--- a/cmd/subctl/gcp.go
+++ b/cmd/subctl/gcp.go
@@ -89,6 +89,9 @@ func init() {
 	gcpPrepareCmd.Flags().BoolVar(&gcpConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = gcpPrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
+
 	cloudPrepareCmd.AddCommand(gcpPrepareCmd)
 
 	addGCPGeneralFlags(gcpCleanupCmd)
--- a/cmd/subctl/rhos.go
+++ b/cmd/subctl/rhos.go
@@ -81,6 +81,9 @@ func init() {
 	rhosPrepareCmd.Flags().BoolVar(&rhosConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = rhosPrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
+
 	cloudPrepareCmd.AddCommand(rhosPrepareCmd)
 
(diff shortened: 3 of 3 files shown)


cmd/subctl/gcp.go
```diff
@@ -89,6 +89,9 @@ func init() {
 	gcpPrepareCmd.Flags().BoolVar(&gcpConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = azurePrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "Deprecated in 0.15, to be removed in 0.16."+
```
sridhargaddam: azurePrepareCmd?

cmd/subctl/rhos.go
```diff
@@ -81,6 +81,9 @@ func init() {
 	rhosPrepareCmd.Flags().BoolVar(&rhosConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = azurePrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "Deprecated in 0.15, to be removed in 0.16."+
```
sridhargaddam: azurePrepareCmd?

sridhargaddam (changes requested):

cmd/subctl/gcp.go
mkolesnik: bad copy paste job :astonished:

cmd/subctl/azure.go
```diff
@@ -79,6 +79,9 @@ func init() {
 		"Whether a dedicated gateway node has to be deployed")
 	cloudPrepareCmd.AddCommand(azurePrepareCmd)
 
+	_ = azurePrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
```
skitt, line 83:
```suggestion
		"To deploy without dedicated gateways, use Load Balancer mode instead.")
```

cmd/subctl/gcp.go
```diff
@@ -89,6 +89,9 @@ func init() {
 	gcpPrepareCmd.Flags().BoolVar(&gcpConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = gcpPrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
```
skitt, line 93:
```suggestion
		"To deploy without dedicated gateways, use Load Balancer mode instead.")
```

cmd/subctl/rhos.go
```diff
@@ -81,6 +81,9 @@ func init() {
 	rhosPrepareCmd.Flags().BoolVar(&rhosConfig.DedicatedGateway, "dedicated-gateway", true,
 		"Whether a dedicated gateway node has to be deployed")
 
+	_ = rhosPrepareCmd.Flags().MarkDeprecated("dedicated-gateway", "to be removed in 0.16. "+
+		"To deploy without dedicated gateways, use the Load Balancer mode instead.")
```
skitt, line 85:
```suggestion
		"To deploy without dedicated gateways, use Load Balancer mode instead.")
```

skitt (approved):
No need to respin just for this, especially since it’s a deprecation message which will only be included in a single release.

tpantelis (approved):

sridhargaddam (approved):

PR #588, opened 2023-03-02
Refactor RecoverData to rely on WriteInfoToFile

skitt:
Depends on https://github.com/submariner-io/subctl/pull/467

### DIFF
--- a/pkg/broker/file.go
+++ b/pkg/broker/file.go
@@ -38,24 +38,11 @@ import (
 
 const InfoFileName = "broker-info.subm"
 
-func WriteInfoToFile(restConfig *rest.Config, brokerNamespace, ipsecFile string, components sets.Set[string],
+func WriteInfoToFile(restConfig *rest.Config, brokerNamespace string, ipsecPSK []byte, components sets.Set[string],
 	customDomains []string, status reporter.Interface,
 ) error {
 	status.Start("Saving broker info to file %q", InfoFileName)
 	defer status.End()
-
-	kubeClient, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		return status.Error(err, "error creating Kubernetes client")
-	}
-
-	data, err := newDataFrom(kubeClient, brokerNamespace, ipsecFile)
-	if err != nil {
-		// TODO return reporter.Error(err, "error initializing broker info")
-		return err
-	}
-
-	data.BrokerURL = restConfig.Host + restConfig.APIPath
 
 	newFilename, err := backupIfExists(InfoFileName)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


dfarrell07 (approved):

Jaanki (approved):

pkg/broker/file.go
```diff
@@ -101,20 +101,17 @@ func newDataFrom(kubeClient kubernetes.Interface, brokerNamespace, ipsecFile str
...
 
-	if ipsecFile != "" {
-		ipsecData, err := ReadInfoFromFile(ipsecFile)
+	psk := ipsecPSK
+	if len(psk) == 0 {
+		psk, err = generateRandomPSK(ipsecSecretLength)
 		if err != nil {
-			return nil, errors.Wrapf(err, "error importing IPsec PSK from file %q", ipsecFile)
+			return nil, err
 		}
```
mkolesnik: I think this logic needs to be outside the scope of `newDataFrom` and `WriteInfoToFile`, it should be part of `deployBrokerInContext` right after trying to read data from file.

E.G. `WriteInfoToFile` should do what it says and not decide if missing info should be added or not, that decision should happen when calling it (the recover command shouldn't invent a PSK if it's missing)

pkg/broker/file.go
```diff
@@ -92,7 +92,7 @@ func ReadInfoFromFile(filename string) (*Info, error) {
 	return data, errors.Wrap(json.Unmarshal(bytes, data), "error unmarshalling data")
 }
 
-func newDataFrom(kubeClient kubernetes.Interface, brokerNamespace, ipsecFile string) (*Info, error) {
+func newDataFrom(kubeClient kubernetes.Interface, brokerNamespace string, ipsecPSK []byte) (*Info, error) {
```
mkolesnik: Not critical to the change, but since you're already touching it you could refactor the function to receive the `restConfig` and create `kubeClient` inside the function, as it's the only one using it.

Looking at it a bit further, the whole function is silly and could be broken down into "read client token" and "initialize PSK" which both can be called from `WriteInfoToFile`.

mkolesnik (changes requested):
Thanks Stephen, this is what I was talking about.

I have only one comment about the logic there.

pkg/broker/file.go
tpantelis: > Not critical to the change, but since you're already touching it you could refactor the function to receive the `restConfig` and create `kubeClient` inside the function, as it's the only one using it.

Keeping it as `kubeClient` will make it easier to add unit tests in the future.

mkolesnik: As I wrote this function is quite silly and can just be removed overall, even more so if we agree on https://github.com/submariner-io/subctl/pull/588/files#r1129189996

Let's see what @skitt thinks

mkolesnik (approved):
Thanks!

PR #589, opened 2023-03-03
Make getSpecBool generic, simplify its return

skitt:
getSpecBool still referred to the role it had before it was extracted as a separate function, with an isDisabled variable. Generify it by naming the variable "value", in the same style as getSpecString.

Also simplify the return statement and clarify the test on found.


dfarrell07 (approved):

tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
[Subctl Commands](https://github.com/submariner-io/subctl/actions/runs/4344058951/jobs/7586877511#logs) failed, rerunning.

PR #621, opened 2023-03-24
Use only subm cluster to recover broker-info.subm

Jaanki:
Epic: https://github.com/submariner-io/enhancements/issues/143

### DIFF
--- a/cmd/subctl/recover_broker_info.go
+++ b/cmd/subctl/recover_broker_info.go
@@ -19,25 +19,26 @@ limitations under the License.
 package subctl
 
 import (
-	"errors"
+	"context"
 	"fmt"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/subctl/internal/cli"
-	"github.com/submariner-io/subctl/internal/constants"
 	"github.com/submariner-io/subctl/internal/exit"
 	"github.com/submariner-io/subctl/internal/restconfig"
 	"github.com/submariner-io/subctl/pkg/broker"
+	"github.com/submariner-io/subctl/pkg/brokercr"
+	"github.com/submariner-io/subctl/pkg/client"
 	"github.com/submariner-io/subctl/pkg/cluster"
 	"github.com/submariner-io/submariner-operator/api/v1alpha1"
-	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/rest"
+	controllerClient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var recoverRestConfigProducer = restconfig.NewProducer().WithPrefixedContext("broker").
-	WithDefaultNamespace(constants.DefaultBrokerNamespace).
-	WithPrefixedNamespace("broker", constants.DefaultBrokerNamespace)
+var recoverRestConfigProducer = restconfig.NewProducer()
 
 // recoverBrokerInfo represents the reconstruct command.
 var recoverBrokerInfo = &cobra.Command{
@@ -45,16 +46,8 @@ var recoverBrokerInfo = &cobra.Command{
 	Short: "Recovers the broker-info.subm file from the installed Broker",
 	Run: func(cmd *cobra.Command, args []string) {
 		status := cli.NewReporter()
-		// if --brokerconfig flag provided, get the broker and proceed to get Submariner
-		contextFound, err := recoverRestConfigProducer.RunOnSelectedPrefixedContext("broker",
-			recoverBrokerFromConfigContext, status)
-		exit.OnError(err)
 
-		// if --brokerconfig not provided, search for broker on current context and proceed to get Submariner
-		if !contextFound {
-			err = recoverRestConfigProducer.RunOnSelectedContext(recoverBrokerFromCurrentContext, status)
-		}
-		exit.OnError(err)
+		exit.OnError(recoverRestConfigProducer.RunOnSelectedContext(restconfig.IfConnectivityInstalled(recoverBrokerInfoFromSubm), status))
 	},
 }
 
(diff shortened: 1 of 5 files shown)


cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
+	brokerNamespace := submCluster.Submariner.Spec.BrokerK8sRemoteNamespace
+	brokerRestConfig := submCluster.RestConfig
 
-	clusters, err := brokerCluster.GetClusters(brokerNamespace)
+	status.Start("Checking if the Broker is installed on the Submariner cluster %q in namespace %q", submCluster.Name, brokerNamespace)
+
+	brokerObj, err := getBroker(brokerRestConfig, brokerNamespace)
 	if err != nil {
-		return status.Error(err, "error listing joined clusters")
+		status.Warning("Error getting Broker: %s", err)
```
tpantelis: An error should be logged as such and abort the operation.

```suggestion
		return status.Error(err, "")
```

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
 	}
 
-	ok, err := tryToRecoverFromBroker(brokerCluster, brokerObj, brokerNamespace, clusters, status)
-	if ok || err != nil {
-		return err
+	if brokerObj == nil {
+		status.Start("Trying to connect to the Broker installed on a different cluster")
+
+		brokerRestConfig, brokerNamespace, err = restconfig.ForBroker(submCluster.Submariner, submCluster.ServiceDiscovery)
+		if err != nil || brokerRestConfig == nil {
```
tpantelis: `brokerRestConfig` won't be nil if there's no error.

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
+			return status.Error(err, "Error getting the broker's rest config")
```
tpantelis:
```suggestion
			return status.Error(err, "Error getting the Broker's REST config")
```

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
+		}
+
+		brokerObj, err = getBroker(brokerRestConfig, brokerNamespace)
+		if err != nil {
+			return status.Error(err, "Error getting Broker")
```
tpantelis: The error already contains enough info.

```suggestion
			return status.Error(err, "")
```

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
+		}, brokerObj)
 
-func getBroker(clusterInfo *cluster.Info, namespace, notFoundMsg string, status reporter.Interface) (*v1alpha1.Broker, error) {
-	status.Start("Checking if the Broker is installed on cluster %q in namespace %q", clusterInfo.Name, namespace)
-
-	brokerObj, err := clusterInfo.GetBroker(namespace)
 	if apierrors.IsNotFound(err) {
-		return nil, status.Error(fmt.Errorf("the Broker is not installed on the specified cluster in namespace %s. %s", notFoundMsg,
-			namespace), "")
+		return nil, errors.Wrapf(err, "the Broker is not installed on the specified cluster in namespace %s", namespace)
```
tpantelis: This is not an error condition. We should return nil error. The `nilnil` linter may complain about this in which case return a `found` bool.

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
+
+		brokerRestConfig, brokerNamespace, err = restconfig.ForBroker(submCluster.Submariner, submCluster.ServiceDiscovery)
+		if err != nil || brokerRestConfig == nil {
+			return status.Error(err, "Error getting the broker's rest config")
+		}
+
+		brokerObj, err = getBroker(brokerRestConfig, brokerNamespace)
+		if err != nil {
+			return status.Error(err, "Error getting Broker")
+		}
```
tpantelis, line 82: We should also check if the `brokerObj` isn't found. ACM doesn't create a `Broker` resource and neither does helm I believe (currently anyway).

```suggestion
		}
		
		if brokerObj == nil {
		    return status.Error(fmt.Errorf("No Broker resource found in namespace %q", brokerNamespace), "")
		}
```

cmd/subctl/recover_broker_info.go
```diff
@@ -63,114 +55,57 @@ func init() {
...
 	}
 
-	status.Warning("Submariner is not installed on the same cluster as Broker")
-	status.Start("Checking if Submariner is installed on a different cluster")
+	status.Success("Found Broker installed on a different cluster in namespace %s", brokerNamespace)
```
tpantelis: This should be inside the `if` block above. Also we need to end the status here.

```suggestion
	status.End()
```

tpantelis (changes requested):

cmd/subctl/recover_broker_info.go
Jaanki: We don't want to return here. This addresses the case where Broker is not installed on the current context and hence we move ahead to getting remote Brokerconfig and get BrokerObj from that remote cluster.

cmd/subctl/recover_broker_info.go
Jaanki: We don't intent to use `recover-broker-info` command with ACM anyways. However, I will make the suggested changes.

cmd/subctl/recover_broker_info.go
Jaanki: All other usages have lower cases. Hence I used lower case here. Anyways, I will change it.

cmd/subctl/recover_broker_info.go
Jaanki: Changed the return err in L103 to be able to return here.

cmd/subctl/recover_broker_info.go
tpantelis: Right but we can't control what a user might do 😄

cmd/subctl/recover_broker_info.go
tpantelis: yeah this goes along with my comment in `getBroker` - I should've been clear about that. Basically we need to distinguish between "not found" and an unrecoverable error and, for the latter, abort. You could also just have `getBroker` propagate the error as is and check `apierrors.IsNotFound` here.

cmd/subctl/recover_broker_info.go
tpantelis: yeah it's fine either way. We should probably be consistent with "Broker" vs "broker" even tho this is an error message and won't normally be seen.

cmd/subctl/recover_broker_info.go
tpantelis: An alternative is to propagate the error as is and check `apierrors.IsNotFound` in the caller. I think that's more idiomatic.

tpantelis (approved):

maayanf24 (approved):

PR #622, opened 2023-03-24
Skip pod-security label checks for the submariner-operator NS

tpantelis:
The pod-security warning was observed on an OCP 4.10.10 cluster when running `subctl diagnose` for the **submariner-operator** NS b/c the pod-security label wasn't explicitly added to the `Namespace` resource. However, the SCC is properly set up for the **submariner-operator** NS so the pod-security label isn't needed. The label check was added in case the user specifies a custom NS for the diagnose tests so skip the check otherwise.


sridhargaddam:
> However, the SCC is properly set up for the submariner-operator NS so the pod-security label isn't needed.

@tpantelis Can you point me to the code where this is setup?

tpantelis:
> > However, the SCC is properly set up for the submariner-operator NS so the pod-security label isn't needed.
> 
> @tpantelis Can you point me to the code where this is setup?

https://github.com/submariner-io/submariner-operator/blob/devel/config/rbac/submariner-operator/ocp_cluster_role.yaml

sridhargaddam:
> However, the SCC is properly set up for the submariner-operator NS so the pod-security label isn't needed.

@tpantelis Can you point me to the code where this is setup?

sridhargaddam:
> However, the SCC is properly set up for the submariner-operator NS so the pod-security label isn't needed.

@tpantelis Can you point me to the code where this is setup?

internal/pods/schedule.go
```diff
@@ -300,6 +300,10 @@ func addNodeSelectorTerm(nodeSelTerms []v1.NodeSelectorTerm, label string,
 }
 
 func checkNSLabels(config *Config) error {
+	if config.Namespace == constants.OperatorNamespace {
```
mkolesnik, line 303: Would be good to have a comment explaining the "why" here, so that future people reading the code understand why it's skipped on this namespace

sridhargaddam (approved):

nyechiel (approved):

PR #627, opened 2023-03-28
Update subctl diagnose service-discovery

tpantelis:
...for the `ServiceImport` aggregation.

- Retain the check for the local `ServiceImport` in the **submariner-operator** NS.

- If the local `ServiceImport` contains the MCS-defined service name label, then the cluster has been upgraded to 0.15, in which case check for the aggregated `ServiceImport` in the service NS.

- Verify the `Conflict` status isn't present in the `ServiceExport`.

Depends on https://github.com/submariner-io/submariner-operator/pull/2551

### DIFF
--- a/pkg/diagnose/servicediscovery.go
+++ b/pkg/diagnose/servicediscovery.go
@@ -20,11 +20,11 @@ package diagnose
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/reporter"
+	"github.com/submariner-io/admiral/pkg/resource"
 	lhconstants "github.com/submariner-io/lighthouse/pkg/constants"
 	"github.com/submariner-io/subctl/internal/constants"
 	"github.com/submariner-io/subctl/internal/gvr"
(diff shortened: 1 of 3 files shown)


pkg/diagnose/servicediscovery.go
```diff
@@ -98,32 +99,51 @@ func checkServiceExport(clusterInfo *cluster.Info, status reporter.Interface) {
 			status.Failure("No EndpointSlice found for exported service %s/%s", se.Namespace, se.Name)
 		}
 
-		_, err = clusterInfo.ClientProducer.ForDynamic().Resource(serviceImportsGVR).
+		checkForAggregateSI := false
+
+		localSI, err := clusterInfo.ClientProducer.ForDynamic().Resource(serviceImportsGVR).
```
skitt: This would be easier to read if `clusterInfo.ClientProducer.ForDynamic().Resource(serviceImportsGVR)` was stored in a separate variable (and it would be re-usable later on).

skitt (approved):
A minor tweak if you respin.

dfarrell07 (approved):

aswinsuryan (approved):

PR #645, opened 2023-04-03
Support for manual trigger of release image job

sridhargaddam:
We have seen instances where the release job does not get triggered after a PR gets merged and because of this, the devel image still points to some old commit which produces inconsistent results. This PR includes support to manually trigger the release job in such scenarios.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,7 @@
 name: Release Images
 
 on:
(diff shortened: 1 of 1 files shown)


skitt (approved):

nyechiel (approved):

skitt:
It turns out we need to backport this for it to be usable on release branches...

PR #649, opened 2023-04-04
Print broker info subm file when not same

Jaanki:
When the recovered and original broker-info.subm files don't match, it
is difficult to debug in the CI.

This PR prints both the file contents in case they don't match.

### DIFF
--- a/scripts/test/system.sh
+++ b/scripts/test/system.sh
@@ -110,6 +110,30 @@ EOF
     echo "::endgroup::"
 }
 
+function test_subctl_recover_broker_info() {
+    # Rename old broker-info.subm to be able to compare it with the recovered broker-info.sum
+    echo "Renaming broker-info.subm to broker-info.subm.orig"
+    mv "${DAPPER_SOURCE}"/output/broker-info.subm "${DAPPER_SOURCE}"/output/broker-info.subm.orig
+    base64 -d "$DAPPER_SOURCE"/output/broker-info.subm.orig > "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig
+
+    _subctl recover-broker-info --context cluster1
+    validate_and_clean_broker_info
+
+    _subctl recover-broker-info --context cluster2
+    validate_and_clean_broker_info
+}
+
+function validate_and_clean_broker_info() {
+  base64 -d broker-info.subm > decoded_broker_info.subm
+  if ! diff <(yq -P eval decoded_broker_info.subm) <(yq -P eval "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig); then
+    echo "Printing the original broker_info.subm file"
+    cat decoded_broker_info.subm.orig
+    echo "Printing the recovered broker_info.subm file"
+    cat decoded_broker_info.subm
+  fi
+  rm -f broker-info.subm decoded_broker_info.subm
+}
+
 ### Main ###
 
 subm_ns=submariner-operator
@@ -176,18 +200,8 @@ _subctl cloud prepare generic --context cluster1
 # Deprecated variant
 _subctl cloud prepare generic --kubecontext cluster1 && exit 1
 
-# Rename old broker-info.subm to be able to compare it with the recovered broker-info.sum
-echo "Renaming broker-info.subm to broker-info.subm.orig"
-mv "${DAPPER_SOURCE}"/output/broker-info.subm "${DAPPER_SOURCE}"/output/broker-info.subm.orig
-
 # Test subctl recover-broker-info invocations
-_subctl recover-broker-info --context cluster1
-cmp "${DAPPER_SOURCE}"/output/broker-info.subm.orig broker-info.subm
-rm -f broker-info.subm
-
-_subctl recover-broker-info --context cluster2
-cmp "${DAPPER_SOURCE}"/output/broker-info.subm.orig broker-info.subm
-rm -f broker-info.subm
+test_subctl_recover_broker_info
 
 # Test subctl uninstall invocations
 _subctl uninstall -y --context cluster2
--- a/pkg/broker/file.go
+++ b/pkg/broker/file.go
@@ -23,6 +23,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


scripts/test/system.sh
```diff
@@ -110,6 +110,12 @@ EOF
     echo "::endgroup::"
 }
 
+function cmp_rm_files() {
+  base64 -d broker-info.subm > decoded_broker_info.subm
+  diff decoded_broker_info.subm "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig
```
mkolesnik: Better to pass through `yq` in order to avoid errors due to ordering: https://mikefarah.gitbook.io/yq/usage/tips-and-tricks#comparing-yaml-files

scripts/test/system.sh
```diff
@@ -179,15 +185,14 @@ _subctl cloud prepare generic --kubecontext cluster1 && exit 1
 # Rename old broker-info.subm to be able to compare it with the recovered broker-info.sum
 echo "Renaming broker-info.subm to broker-info.subm.orig"
 mv "${DAPPER_SOURCE}"/output/broker-info.subm "${DAPPER_SOURCE}"/output/broker-info.subm.orig
+base64 -d "$DAPPER_SOURCE"/output/broker-info.subm.orig > "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig
 
 # Test subctl recover-broker-info invocations
 _subctl recover-broker-info --context cluster1
-cmp "${DAPPER_SOURCE}"/output/broker-info.subm.orig broker-info.subm
-rm -f broker-info.subm
+cmp_rm_files
```
mkolesnik: I'd combine it all to a single command `test_subctl_recover_broker_info` or some such name

Then moving old `broker-info.subm` and decoding it can be hidden inside as well, as it only pertains to this test case, making tests clearer

mkolesnik (changes requested):

scripts/test/system.sh
Jaanki: sort_keys sorts the `data` struct based on the keys. But since the order of keys would always remain same (ref [this stack overflow thread](https://stackoverflow.com/questions/68404867/how-do-i-change-the-position-of-golang-struct-values/68405135#68405135)) and we do [use `json.Marshal`](https://github.com/submariner-io/subctl/blob/devel/pkg/broker/info.go#L61) while writing to `broker-info.subm` file, I think its not necessary. I will use yq to pretty-print the output.

scripts/test/system.sh
```diff
@@ -110,6 +110,25 @@ EOF
...
+
+    _subctl recover-broker-info --context cluster2
+    cmp_rm_files
+}
+
+function cmp_rm_files() {
+  base64 -d broker-info.subm > decoded_broker_info.subm
+  diff <(yq -P eval decoded_broker_info.subm) <(yq -P eval "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig)
+  rm -f broker-info.subm decoded_broker_info.subm
+}
```
mkolesnik: What I mean is something like this:

```suggestion
function test_subctl_recover_broker_info() {
    # Rename old broker-info.subm to be able to compare it with the recovered broker-info.sum
    echo "Renaming broker-info.subm to broker-info.subm.orig"
    mv "${DAPPER_SOURCE}"/output/broker-info.subm "${DAPPER_SOURCE}"/output/broker-info.subm.orig
    base64 -d "$DAPPER_SOURCE"/output/broker-info.subm.orig > "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig

  _subctl recover-broker-info "$@"
  base64 -d broker-info.subm > decoded_broker_info.subm
  diff <(yq -P eval decoded_broker_info.subm) <(yq -P eval "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig)
  rm -f broker-info.subm decoded_broker_info.subm
  
  mv "${DAPPER_SOURCE}"/output/broker-info.subm.orig "${DAPPER_SOURCE}"/output/broker-info.subm
}

... later in the file ...

test_subctl_recover_broker_info --context cluster1

test_subctl_recover_broker_info --context cluster2
```

mkolesnik (changes requested):

scripts/test/system.sh
Jaanki: Renaming and decoding the old `broker-info.subm` should only be done. No need to do that for each command call. What you are suggesting is not expected behaviur. The recovered `broker-info.subm` needs to be created and decoded on each command run.

mkolesnik: We're talking about system tests here, yes?

What's your concern exactly? Performance?

I prefer each test to be self contained as possible and take care of it's set up and clean up so that other tests can run after it.
Having it all in the function serves this purpose very well, while having a small "performance" impact which is negligible since these are tests, not code running in a production system.

mkolesnik (commented):
As I'm going away for holiday shutdown I won't be reviewing this any time soon, so changing my vote to a "comment" to not hold back this patch.

Will let others weigh in on my comments and follow/reject as they see fit.

mkolesnik (dismissed):

scripts/test/system.sh
tpantelis: Either way is fine - as Mike said, it's just a system test - but Mike's approach seems a bit cleaner, ie it isolates the code in one function.

scripts/test/system.sh
```diff
@@ -110,6 +110,25 @@ EOF
...
+    base64 -d "$DAPPER_SOURCE"/output/broker-info.subm.orig > "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig
+
+    _subctl recover-broker-info --context cluster1
+    cmp_rm_files
+
+    _subctl recover-broker-info --context cluster2
+    cmp_rm_files
+}
+
+function cmp_rm_files() {
```
skitt: If we keep this, it should have a name indicating that it’s specific to `broker-info.subm`, and indicates its purpose rather than its implementation, _e.g._ `validate_and_clean_broker_info`.

skitt (changes requested):

scripts/test/system.sh
skitt: This follows the model of `subctl_gather`.

skitt:
To avoid comparison errors based on the order of components, the components should be sorted in `WriteInfoToFile`, _e.g._ using `sort.Strings()`.

scripts/test/system.sh
Jaanki: > This follows the model of `subctl_gather`.

and `test_subctl_diagnose_in_cluster`

skitt (approved):

tpantelis (approved):

Jaanki:
Backport: https://github.com/submariner-io/subctl/pull/663

PR #653, opened 2023-04-06
Do not create default broker client SA and its RBAC

Jaanki:
When deploying broker, `submariner-k8s-broker-client` SA is created and bound to `submariner-k8s-broker-cluster`. This SA is not used anywhere.

This PR removes this SA.


pkg/broker/rbac.go
```diff
@@ -29,7 +29,6 @@ import (
 const (
 	submarinerBrokerClusterRole      = "submariner-k8s-broker-cluster"
 	submarinerBrokerAdminRole        = "submariner-k8s-broker-admin"
```
skitt:
```suggestion
	submarinerBrokerClusterRole = "submariner-k8s-broker-cluster"
	submarinerBrokerAdminRole   = "submariner-k8s-broker-admin"
```

skitt (changes requested):

sridhargaddam (approved):

skitt (approved):

dfarrell07:
Should this be backported @Jaanki?

tpantelis:
This PR is causing a [system test failure](https://github.com/submariner-io/submariner-operator/actions/runs/4729438042/jobs/8402138154?pr=2601#step:3:1309) in **submariner-operator**. It uses the `submariner-k8s-broker-client` SA/secret to obtain [SUBMARINER_BROKER_CA](https://github.com/submariner-io/submariner-operator/blob/7134634809a8b0407e9ae6198d5fe331c3009a05/scripts/test/system.sh#L16). It's also referenced in [submariner-charts](https://github.com/submariner-io/submariner-charts/blob/61b7d37f76cf19bc81043e0c77fc734696ff244a/submariner-k8s-broker/templates/NOTES.txt#L9). Perhaps we were premature in removing it or the uses/references elsewhere need to removed/updated? @Jaanki @skitt

skitt:
Let’s revert it for now.

skitt:
We should have other SAs that can be used for those purposes, let’s update the test and chart docs to use one of those.

Jaanki:
> Should this be backported @Jaanki?

Not needed @dfarrell07. This is a part of cleanup/sycn I am doing between subctl and operator to use same RBAC definitions defined in operator.

Jaanki:
> We should have other SAs that can be used for those purposes, let’s update the test and chart docs to use one of those.

operator PR: https://github.com/submariner-io/submariner-operator/pull/2602

PR #663, opened 2023-04-11
[Backport] Print broker info subm file when not same

Jaanki:
When the recovered and original broker-info.subm files don't match, it is difficult to debug in the CI.

This PR prints both the file contents in case they don't match.

The reason of mismatch is at times sequence change of `components`.

This PR also sorts the `components` before writing it to broker-info.sybm file so that the sequence remains same always.

Backport of https://github.com/submariner-io/subctl/pull/649

### DIFF
--- a/scripts/test/system.sh
+++ b/scripts/test/system.sh
@@ -110,6 +110,30 @@ EOF
     echo "::endgroup::"
 }
 
+function test_subctl_recover_broker_info() {
+    # Rename old broker-info.subm to be able to compare it with the recovered broker-info.sum
+    echo "Renaming broker-info.subm to broker-info.subm.orig"
+    mv "${DAPPER_SOURCE}"/output/broker-info.subm "${DAPPER_SOURCE}"/output/broker-info.subm.orig
+    base64 -d "$DAPPER_SOURCE"/output/broker-info.subm.orig > "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig
+
+    _subctl recover-broker-info --context cluster1
+    validate_and_clean_broker_info
+
+    _subctl recover-broker-info --context cluster2
+    validate_and_clean_broker_info
+}
+
+function validate_and_clean_broker_info() {
+  base64 -d broker-info.subm > decoded_broker_info.subm
+  if ! diff <(yq -P eval decoded_broker_info.subm) <(yq -P eval "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig); then
+    echo "Printing the original broker_info.subm file"
+    cat decoded_broker_info.subm.orig
+    echo "Printing the recovered broker_info.subm file"
+    cat decoded_broker_info.subm
+  fi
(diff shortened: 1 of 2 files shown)


scripts/test/system.sh
```diff
@@ -110,6 +110,30 @@ EOF
...
+    _subctl recover-broker-info --context cluster1
+    validate_and_clean_broker_info
+
+    _subctl recover-broker-info --context cluster2
+    validate_and_clean_broker_info
+}
+
+function validate_and_clean_broker_info() {
+  base64 -d broker-info.subm > decoded_broker_info.subm
+  if ! diff <(yq -P eval decoded_broker_info.subm) <(yq -P eval "$DAPPER_SOURCE"/output/decoded_broker_info.subm.orig); then
```
sridhargaddam, line 130: If the components are not listed in the same order, its still okay from Submariner POV and should not be treated as an error. 

Apart from components not matching, what are the other situations when the contents may not match?

sridhargaddam (changes requested):

scripts/test/system.sh
Jaanki: > If the components are not listed in the same order, its still okay from Submariner POV and should not be treated as an error.
> 
Yes, but for `diff` it matters. So even if both (original and recovered) broker-info.subm files have exact same content, but the order of `components` is different, `diff` fails and so the system test. This PR makes sure the components are written to broker-info.subm in order so that the order is preserved while recovering and `diff` does not fail.

The components are [originally saved ](https://github.com/submariner-io/subctl/blob/devel/pkg/broker/file.go#L74)as `UnsortedList`. With this PR, I am sorting it before storing for diff's sake.

> Apart from components not matching, what are the other situations when the contents may not match?
broker-info.subm has:

```
type Info struct {
	BrokerURL        string         `json:"brokerURL"`
	ClientToken      *corev1.Secret `omitempty,json:"clientToken"`
	IPSecPSK         *corev1.Secret `omitempty,json:"ipsecPSK"`
	ServiceDiscovery bool           `omitempty,json:"serviceDiscovery"`
	Components       []string       `json:",omitempty"`
	CustomDomains    *[]string      `omitempty,json:"customDomains"`
}
```

diff would fail in below 3 scenarios:
1. If the value of any of the fields don't match. Then its a bug.
2. The sequence of fields stored. We won't have this issue because the sequence will always be preserved as explained in [this stack overflow thread](https://stackoverflow.com/questions/68404867/how-do-i-change-the-position-of-golang-struct-values/68405135#68405135)) and we do [use json.Marshal](https://github.com/submariner-io/subctl/blob/devel/pkg/broker/info.go#L61) while writing to broker-info.subm file.
3. The sequence of `components` is not same. Which this PR solves.

Jaanki:
This is a backport of https://github.com/submariner-io/subctl/pull/649

sridhargaddam (approved):

vthapar (approved):

PR #683, opened 2023-04-19
Remove useless empty list check

jinglina:

### DIFF
--- a/cmd/subctl/export.go
+++ b/cmd/subctl/export.go
@@ -66,7 +66,7 @@ func init() {
 }
 
 func validateArguments(args []string) error {
-	if len(args) == 0 || args[0] == "" {
+	if len(args) == 0 {
 		return errors.New("name of the Service to be exported must be specified")
 	}
 


jinglina:
remove useless empty list check

tpantelis (approved):

skitt (changes requested):
This isn’t a useless empty list check: it checks that, when the list of arguments _isn’t_ empty, the first argument isn’t itself empty. Try

    subctl export service ""

You’d never run a command like that intentionally but it can happen if variable expansion is involved.

jinglina:
> This isn’t a useless empty list check: it checks that, when the list of arguments _isn’t_ empty, the first argument isn’t itself empty. Try
> 
> ```
> subctl export service ""
> ```
> 
> You’d never run a command like that intentionally but it can happen if variable expansion is involved.

This is beyond my testing. Thanks. Then I think `subctl join` also need this check  , right?

skitt:
> This is beyond my testing. Thanks. Then I think `subctl join` also need this check , right?

I’m not sure whether it makes sense to keep the test, I’m checking other tools’ behaviour...

skitt (approved):
OK, applying this results in the same behaviour as `kubectl`, so it’s fine by me.

PR #747, opened 2023-05-28
WIP: Read gather version from submariner.status

### LINKED ISSUE #746: Subctl gather should read the Submariner version from status field instead of spec
**What happened**:

Currently when `subctl gather ...` is executed against a cluster, it tries to get the Submariner version from the `submariners.spec.version` field. However, it should read this info from `submariners.status.version` as there is some discrepancy observed when a different version of subctl is used which does not match with the installed version.

yboaron:
Retrieve subctl gather version from submariner.status instead of submariner.spec

Fixes:  https://github.com/submariner-io/subctl/issues/746

### DIFF
--- a/internal/gather/summary.go
+++ b/internal/gather/summary.go
@@ -123,7 +123,7 @@ func getVersions(info *Info) version {
 
 	Versions.Subm = "Not installed"
 	if info.Submariner != nil {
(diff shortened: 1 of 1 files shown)


yboaron:
I'm going to close this PR

PR #794, opened 2023-06-15
Add build constraints to allow buildling subctl with only non-deploy …

maayanf24:
…commands

Related to: #https://github.com/submariner-io/enhancements/issues/196


pkg/cloud/aws/aws.go
```diff
@@ -1,3 +1,5 @@
+//go:build !non_deploy
```
Jaanki: No need to add constraints in this file. Applies to all files under pkg/. Only files under cmd/subctl should have the constraint.

Jaanki (changes requested):
Constraining only the cobra related files will do the trick. i.e add constraints only to files under `cmd/subctl/*` and not `pkg/*`

skitt (changes requested):
Please add a system test: build the non-deploy `subctl`, and use it to run `show`, `diagnose` etc.

Please also shorten the commit message:

> Add build constraints for a non-deploy subctl

maayanf24:
> Please add a system test: build the non-deploy `subctl`, and use it to run `show`, `diagnose` etc.
> 
> Please also shorten the commit message:
> 
> > Add build constraints for a non-deploy subctl

After discussing with @Jaanki I only added test for `non_deploy_subctl show all` to save the time it would take to also run `diagnose`, `gather` and the other commands

skitt:
> After discussing with @Jaanki I only added test for `non_deploy_subctl show all` to save the time it would take to also run `diagnose`, `gather` and the other commands

I see you also added sanity tests to make sure that non-deploy `subctl` really can’t `deploy-broker` and `join`, nice!

skitt (approved):
I’m looking into the linting error.

Makefile
```diff
@@ -125,6 +125,9 @@ cmd/bin/subctl-%: $(shell find . -name "*.go")
 		-X 'github.com/submariner-io/submariner-operator/api/v1alpha1.DefaultRepo=$(DEFAULT_REPO)'" \
 	$(SCRIPTS_DIR)/compile.sh $@ ./cmd
 
+cmd/bin/non_deploy_subctl:
+	$(GO) build -o cmd/bin/non_deploy_subctl --tags non_deploy ./cmd
```
Jaanki: Add LDFLAGS like its done at L123 above.

scripts/test/system.sh
```diff
@@ -198,6 +206,13 @@ _subctl benchmark throughput "${KUBECONFIGS_DIR}"/kind-config-cluster1 "${KUBECO
 
 # Test subctl cloud prepare invocations
 
+# test non deploy subctl for deploy commands
+_non_deploy_subctl deploy-broker --kubeconfig "${KUBECONFIGS_DIR}"/kind-config-cluster1 && exit 1
+_non_deploy_subctl join --kubeconfig "${KUBECONFIGS_DIR}"/kind-config-cluster1 broker-info.subm --clusterid cluster1 && exit 1
```
Jaanki, line 211: Add similar tests for all ignored commands. That way, we don't miss any. In future, if the build constraint is accidentally removed from some commands, this job should fail.

Jaanki (changes requested):

skitt:
The linting error is [an upstream bug](https://github.com/denis-tingaikin/go-header/issues/18), I think the only thing we can do (short of using another tool to check headers) is to ignore files with a build constraint. At the end of `.golangci.yml`, add an `issues.exclude-rules` rule for the `goheaders` linter on the relevant paths.

Makefile
```diff
@@ -125,6 +125,14 @@ cmd/bin/subctl-%: $(shell find . -name "*.go")
 		-X 'github.com/submariner-io/submariner-operator/api/v1alpha1.DefaultRepo=$(DEFAULT_REPO)'" \
 	$(SCRIPTS_DIR)/compile.sh $@ ./cmd
 
+cmd/bin/non_deploy_subctl:
+	mkdir -p cmd/bin
```
skitt: Let’s use some Make variables to avoid repeating stuff:

```suggestion
	mkdir -p $(@D)
```

Makefile
```diff
@@ -125,6 +125,14 @@ cmd/bin/subctl-%: $(shell find . -name "*.go")
...
+	$(GO) build -o cmd/bin/non_deploy_subctl --tags non_deploy \
```
skitt:
```suggestion
	$(GO) build -o $@ --tags non_deploy \
```

.golangci.yml
```diff
@@ -201,3 +201,8 @@ issues:
     - path: internal/(cli|env|log)/.*\.go
       linters:
         - goheader
+
+    # Ignore header linting for build constraint
```
skitt, line 205:
```suggestion
    # Ignore header linting for build constraint
    # See https://github.com/denis-tingaikin/go-header/issues/18
```

skitt (changes requested):
Sorry, a few Makefile tweaks I didn’t think of earlier, and a pointer to the goheader bug pushing us to add golangci-lint exclusions.

Makefile
maayanf24: Should I update L116 as well?

skitt: Ah yes, let’s do that, thanks!

skitt (approved):

skitt:
@Jaanki can you re-review?

Jaanki (approved):

tpantelis (approved):

PR #878, opened 2023-08-17
Show actual running versions

### LINKED ISSUE #798: show versions should extract version information from pod logs

In addition to the `DaemonSet`/`Deployment` version shown by `show versions`, we should also extract version information from pod logs.

**Why is this needed**:

This will make it simpler to determine versions of Submariner components running in a cluster, and will ensure we get accurate information regardless of image overrides, mirrors etc.

Depends on https://github.com/submariner-io/submariner/issues/2548
Depends on https://github.com/submariner-io/lighthouse/issues/1279
Depends on https://github.com/submariner-io/shipyard/issues/1327
Depends on https://github.com/submariner-io/lighthouse/issues/1333
Depends on https://github.com/submariner-io/submariner/issues/2633
Depends on https://github.com/submariner-io/submariner-operator/issues/2751

vthapar:
Modify show versions to show actual running version on the pod, not just configured one.

Fixes #798

### DIFF
--- a/internal/show/versions.go
+++ b/internal/show/versions.go
@@ -19,24 +19,43 @@ limitations under the License.
 package show
 
 import (
+	"bufio"
 	"context"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/names"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/subctl/internal/constants"
+	"github.com/submariner-io/subctl/internal/pods"
 	"github.com/submariner-io/subctl/internal/show/table"
 	"github.com/submariner-io/subctl/pkg/cluster"
 	"github.com/submariner-io/submariner-operator/pkg/images"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
+
+const maxLogLinesToScan = 20
+
+var componentCmd = map[string][]string{
+	names.RouteAgentComponent:        {"submariner-route-agent", "--version"},
+	names.GatewayComponent:           {"submariner-gateway", "--version"},
+	names.GlobalnetComponent:         {"submariner-globalnet", "--version"},
+	names.ServiceDiscoveryComponent:  {"lighthouse-agent", "--version"},
+	names.LighthouseCoreDNSComponent: {"lighthouse-coredns", "--subm-version"},
+	names.OperatorComponent:          {"submariner-operator", "--version"},
+	names.MetricsProxyComponent:      {"cat", "version"},
+}
 
 func printDaemonSetVersions(clusterInfo *cluster.Info, printer *table.Printer, components ...string) error {
 	daemonSets := clusterInfo.ClientProducer.ForKubernetes().AppsV1().DaemonSets(constants.OperatorNamespace)
 
 	for _, component := range components {
-		daemonSet, err := daemonSets.Get(context.TODO(), component, v1.GetOptions{})
+		daemonSet, err := daemonSets.Get(context.TODO(), component, metav1.GetOptions{})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				continue
@@ -47,7 +66,14 @@ func printDaemonSetVersions(clusterInfo *cluster.Info, printer *table.Printer, c
 
 		// The name of the function is confusing, it just parses any image repo & version
 		version, repository := images.ParseOperatorImage(daemonSet.Spec.Template.Spec.Containers[0].Image)
-		printer.Add(component, repository, version)
+
+		runningVersion, err := getVersionForComponent(clusterInfo, component,
+			labels.SelectorFromSet(daemonSet.Spec.Selector.MatchLabels))
+		if err != nil {
+			return errors.Wrapf(err, "error retrieving running version for %s", component)
+		}
+
+		printer.Add(component, repository, version, runningVersion)
 	}
 
 	return nil
@@ -57,7 +83,7 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer, 
 	deployments := clusterInfo.ClientProducer.ForKubernetes().AppsV1().Deployments(constants.OperatorNamespace)
 
 	for _, component := range components {
-		deployment, err := deployments.Get(context.TODO(), component, v1.GetOptions{})
+		deployment, err := deployments.Get(context.TODO(), component, metav1.GetOptions{})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				continue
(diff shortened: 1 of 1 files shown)


internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
 		}
 
 		version, repository := images.ParseOperatorImage(deployment.Spec.Template.Spec.Containers[0].Image)
-		printer.Add(component, repository, version)
+
+		runningVersion, err := getVersionForComponent(clusterInfo, component,
+			labels.SelectorFromSet(deployment.Spec.Selector.MatchLabels).String())
+		if err != nil {
```
vthapar, line 99: Should this be considered an error, or just return version as `Unavailable`? Error handling was useful while working on this change, but throwing an error means `show versions` output can be incomplete or even empty if just one component ran into error.

skitt: Right, it shouldn’t be an error.

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
 
 	return nil
 }
 
+func getVersionForComponent(clusterInfo *cluster.Info, component, labelSelector string) (string, error) {
+	podsClient := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Pods(constants.OperatorNamespace)
+	podList, err := podsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+
+	if err != nil || len(podList.Items) < 1 {
+		return "", errors.Wrapf(err, "Failed to find pods for component %s", component)
```
tpantelis:
```suggestion
		return "", errors.Wrapf(err, "failed to find pods for component %s", component)
```

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+		podVersion = getVersionFromPodLogs(&pod, podsClient, component)
+
+		if podVersion == "" {
+			continue
+		}
+
+		return podVersion, nil
+	}
+
+	return "", errors.Errorf("Failed to get version for %s", component)
```
tpantelis:
```suggestion
	return "", errors.Errorf("could not determine the version for component %q", component)
```

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+			return err
+		}
+
+		printer.Add(component, repository, version, runningVersion)
 	}
 
 	return nil
 }
 
+func getVersionForComponent(clusterInfo *cluster.Info, component, labelSelector string) (string, error) {
```
skitt: Wouldn’t it be better to use the `Selector` type for the label selector, rather than its string form? This would allow selector composition should we ever need that.

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+			return podVersion, nil
+		}
+
+		podVersion = getVersionFromPodLogs(&pod, podsClient, component)
+
+		if podVersion == "" {
+			continue
+		}
+
+		return podVersion, nil
```
skitt: To match the earlier construct,

```suggestion
		if podVersion != "" {
			return podVersion, nil
		}
```

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+
+func getVersionFromPodBinary(pod *corev1.Pod, clusterInfo *cluster.Info, component string) (string, error) {
+	cmdArg := "--version"
+
+	switch component {
+	case names.LighthouseCoreDNSComponent:
+		cmdArg = "--subm-version"
+	case names.MetricsProxyComponent:
+		cmdArg = "version"
+	}
```
skitt: This means we end up encoding information about the commands to use in two places: the map at the start of the file, and here. I think it would be better to specify everything in the map (make it a map of strings to string slices).

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+
+	if component == names.MetricsProxyComponent {
+		return outStr, nil
+	}
+
+	result, found := strings.CutPrefix(errStr, fmt.Sprintf("%s version: ", component))
+
+	if !found {
+		return "", errors.Errorf("Failed to extract %s version from %s", component, errStr)
```
tpantelis: I don't think we should return an error here - an empty version indicates not found.

internal/show/versions.go
```diff
@@ -19,24 +19,44 @@ limitations under the License.
...
+
+var componentCmd = map[string]string{
+	names.NetworkPluginSyncerComponent: "submariner-networkplugin-syncer",
+	names.RouteAgentComponent:          "submariner-route-agent",
+	names.GatewayComponent:             "submariner-gateway",
+	names.GlobalnetComponent:           "submariner-globalnet",
+	names.ServiceDiscoveryComponent:    "lighthouse-agent",
+	names.LighthouseCoreDNSComponent:   "lighthouse-coredns",
+	names.OperatorComponent:            "submariner-operator",
+	names.MetricsProxyComponent:        "cat",
```
skitt: This shouldn’t block this PR, but since we have a specific command for the metrics proxy now, we could add `--version` support to that.

skitt (changes requested):

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+	podsClient := clusterInfo.ClientProducer.ForKubernetes().CoreV1().Pods(constants.OperatorNamespace)
+	podList, err := podsClient.List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector})
+
+	if err != nil || len(podList.Items) < 1 {
+		return "", errors.Wrapf(err, "Failed to find pods for component %s", component)
+	}
+
+	// Try all pods
+	for i := range podList.Items {
+		pod := podList.Items[i]
```
tpantelis:
```suggestion
		pod := &podList.Items[i]
```

internal/show/versions.go
```diff
@@ -67,19 +94,119 @@ func printDeploymentVersions(clusterInfo *cluster.Info, printer *table.Printer,
...
+		podVersion, _ := getVersionFromPodBinary(&pod, clusterInfo, component)
+
+		if podVersion != "" {
+			return podVersion, nil
+		}
```
tpantelis: We shouldn't ignore the return error.

```suggestion
		podVersion, err := getVersionFromPodBinary(&pod, clusterInfo, component)

		if err != nil || podVersion != "" {
			return podVersion, err
		}
```

internal/show/versions.go
tpantelis: How about we make the arg part of `componentCmd`?

```
var componentCmd = map[string][]string{
	names.RouteAgentComponent:                 {"submariner-route-agent", "--version"},
	names.GatewayComponent:                       {"submariner-gateway", "--version"},
	names.GlobalnetComponent:                     {"submariner-globalnet", "--version"},
	names.ServiceDiscoveryComponent:        {"lighthouse-agent", "--version"},
	names.LighthouseCoreDNSComponent:   {"lighthouse-coredns", "--subm-version"},
	names.OperatorComponent:                       {"submariner-operator", "--version"},
	names.MetricsProxyComponent:                {"cat", "version"},
}
```

internal/show/versions.go
```diff
@@ -19,24 +19,44 @@ limitations under the License.
...
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+const maxLogLinesToScan = 20
+
+var componentCmd = map[string]string{
+	names.NetworkPluginSyncerComponent: "submariner-networkplugin-syncer",
```
tpantelis: This is being removed.

tpantelis (dismissed):

internal/show/versions.go
tpantelis: 👍 I suggested that too (our reviews crossed).

internal/show/versions.go
tpantelis: 👍 I suggested that too (our reviews crossed).

internal/show/versions.go
vthapar: Tricky part here is, if we return error `show version` will fail and not show rest of content. Also, in case of multiple pods, first one may have error but next one may still work. From user perspective, it would make sense to not throw error but return "Unavailable" or something like that.

Error would've been useful if subctl had its own log other than the user visible output. Error helps figure out why it failed, but that can be figured out from `gather` output. While error was helpful to me during dev it doesn't seem too helpful for actual users.

internal/show/versions.go
vthapar: I used string coz `ListOptions.LabelSelector` is string.

internal/show/versions.go
tpantelis: A return error should indicate a non-recoverable failure attempting to determine the version, eg K8s API server not available.  If we can't determine a version not due to an actual failure then we shouldn't return an error, instead return "N/A".

internal/show/versions.go
skitt: Right, but that doesn’t mean we can’t keep the higher-level object here, and call `.String()` to provide the value for `LabelSelector`.

internal/show/versions.go
vthapar: We want to ignore error coz we will try getting version from log if failed to fetch from binary.

tpantelis: But an error here should indicate a non-recoverable failure, eg K8s API server not available, whereas if the error hadn't occurred then we would've determined the version. If no version is found then try the pod log. The latter case should only occur for a pre 0.16 version.

skitt (approved):

sridhargaddam (approved):

PR #883, opened 2023-08-22
Remove networkplugin syncer bits

tpantelis:
...from gather, show, image override et al

Depends on https://github.com/submariner-io/submariner-operator/pull/2764

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -11,13 +11,18 @@ jobs:
   e2e:
     name: E2E
     if: contains(github.event.pull_request.labels.*.name, 'ready-to-test') && github.event.pull_request.state == 'open'
-    timeout-minutes: 30
+    timeout-minutes: 120
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
         lighthouse: ['', 'lighthouse']
+        ovn: ['']
+        include:
+          - ovn: 'ovn'
+            lighthouse: 'lighthouse'
+            globalnet: 'globalnet'
     steps:
       - name: Check out the repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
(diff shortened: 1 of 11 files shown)


dfarrell07 (approved):

maayanf24 (approved):

tpantelis:
The delete permissions for the operator aren't quite right - https://github.com/submariner-io/submariner-operator/pull/2764 fixes it. I affected the changes by hardcoding in subctl and the E2E passed.

.github/workflows/e2e-full.yml
```diff
@@ -11,21 +11,22 @@ jobs:
...
     if: contains(github.event.pull_request.labels.*.name, 'ready-to-test') && github.event.pull_request.state == 'open'
-    timeout-minutes: 30
+    timeout-minutes: 120
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         globalnet: ['', 'globalnet']
         lighthouse: ['', 'lighthouse']
+        ovn: ['', 'ovn']
```
sridhargaddam: Continuing from the other PR that was closed.

Reg: https://github.com/submariner-io/subctl/pull/885#discussion_r1304249104

Yes @tpantelis. Since OVN is already validated as part of Submariner repo, AFAIU the main purpose of validating it here is to ensure that any OVN specific changes/dependencies in subctl/operator are verified in these jobs. So, if we validate two jobs, one with GN and the other without GN, I feel it should be fine as this will reduce two jobs.

tpantelis: > if we validate two jobs, one with GN and the other without GN,

I assume you mean two LH jobs in which case we exclude the `OVN` and `OVN + GN` jobs since the **submariner** repo already runs them. I guess that assumes any deployment issue introduced by a **subctl** or dependent **submariner-operator** change related to OVN would be sufficiently covered by just these 2 jobs, ie we wouldn't have a case where the `OVN + LH` job succeeds but `OVN` job doesn't. That sounds reasonable.

tpantelis: I pared it down to just one more job: `OVN + GN + LH`.  I think that should suffice.

sridhargaddam (approved):

.github/workflows/e2e-full.yml
tpantelis: /cc @skitt

skitt (approved):

PR #885, opened 2023-08-23
Add OVN to the E2E test matrix

skitt:

### DIFF
--- a/.github/workflows/e2e-full.yml
+++ b/.github/workflows/e2e-full.yml
@@ -18,6 +18,7 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         lighthouse: ['', 'lighthouse']
+        ovn: ['', 'ovn']
     steps:
       - name: Check out the repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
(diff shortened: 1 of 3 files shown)


maayanf24 (approved):

tpantelis:
Bumping **submariner-operator** version to pull in the RBAC changes also pulls in the embedded yaml changes so we need https://github.com/submariner-io/subctl/pull/883 to fix the build errors. I just included this commit with https://github.com/submariner-io/subctl/pull/883.

.github/workflows/e2e-full.yml
```diff
@@ -18,14 +18,15 @@ jobs:
       matrix:
         globalnet: ['', 'globalnet']
         lighthouse: ['', 'lighthouse']
+        ovn: ['', 'ovn']
```
sridhargaddam, line 21: This looks fine, but as it doubles the number of jobs, can we add two additional jobs using the [include](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) options and validate the following additional jobs?
1. GN + LH + OVN
2. LH + OVN

tpantelis: Note that this commit is now part of https://github.com/submariner-io/subctl/pull/883. This PR adds 4 more jobs, so you're saying you want to exclude these?

1. OVN
2. OVN + GN

tpantelis:
Included with https://github.com/submariner-io/subctl/pull/883

PR #888, opened 2023-08-30
Gather OVN IC resources

vthapar:
Add following resources to `subctl gather` for OVN IC use case
 * GatewayRoutes
 * NonGatewayRoutes

Refer https://github.com/submariner-io/enhancements/issues/186


internal/gather/connectivity.go
```diff
@@ -81,3 +81,11 @@ func gatherGlobalEgressIPs(info *Info) {
 func gatherGlobalIngressIPs(info *Info) {
 	ResourcesToYAMLFile(info, submarinerv1.SchemeGroupVersion.WithResource("globalingressips"), corev1.NamespaceAll, v1.ListOptions{})
 }
+
+func gatherGatewayRoutes(info *Info) {
+	ResourcesToYAMLFile(info, submarinerv1.SchemeGroupVersion.WithResource("gatewayroutes"), corev1.NamespaceAll, v1.ListOptions{})
```
sridhargaddam, line 86: I think these resources are only created in the `submariner-operator` namespace. So, using NamespaceAll may not be required, but there is no harm as such. @aswinsuryan what do you suggest?

vthapar: I went with ALL coz it will help us catch any issues cause dby bugs or users manually creating them in other namespaces. In normal scenarios, it is practically same so data collected will not vary but good to have ALL to catch in case they show up in other namespaces.

skitt (approved):

internal/gather/connectivity.go
aswinsuryan: This resource is similar to the endpoints and clusters that the gateway creates. So I think the gather logic can be similar.

sridhargaddam (approved):

aswinsuryan (approved):

PR #902, opened 2023-09-06
feat: add subctl export namespace support

Xunzhuo:
feat: add subctl export namespace support

provide a way to export all services under a specific namespace:

image

### DIFF
--- a/cmd/subctl/export.go
+++ b/cmd/subctl/export.go
@@ -54,6 +54,21 @@ var (
 				}, cli.NewReporter()))
 		},
 	}
+	exportNamespaceCmd = &cobra.Command{
+		Use:   "namespace <namespaceName>",
+		Short: "Exports Services under a specific namespace to other clusters",
+		Long: "This command creates ServiceExport resources with the given namespace which causes the Services under the namespace of the same name to be accessible" +
+			" to other clusters",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := validateArguments(args)
+			exit.OnErrorWithMessage(err, "Insufficient arguments")
+
+			exit.OnError(exportRestConfigProducer.RunOnSelectedContext(
+				func(clusterInfo *cluster.Info, namespace string, status reporter.Interface) error {
+					return service.Exports(clusterInfo.ClientProducer, args[0], status) //nolint:wrapcheck // No need to wrap errors here.
+				}, cli.NewReporter()))
+		},
+	}
 )
 
 func init() {
@@ -62,6 +77,7 @@ func init() {
 
 	exportRestConfigProducer.SetupFlags(exportServiceCmd.Flags())
 	exportCmd.AddCommand(exportServiceCmd)
+	exportCmd.AddCommand(exportNamespaceCmd)
 	rootCmd.AddCommand(exportCmd)
 }
 
--- a/cmd/subctl/export.go
+++ b/cmd/subctl/export.go
@@ -83,7 +83,7 @@ func init() {
 
 func validateArguments(args []string) error {
 	if len(args) == 0 {
-		return errors.New("name of the Service to be exported must be specified")
+		return errors.New("name of the Service or Namespace to be exported must be specified")
 	}
 
 	return nil
--- a/pkg/service/export.go
+++ b/pkg/service/export.go
@@ -20,6 +20,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/admiral/pkg/resource"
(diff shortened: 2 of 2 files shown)


pkg/service/export.go
```diff
@@ -65,3 +66,46 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
...
+		if err != nil {
+			if k8serrors.IsAlreadyExists(err) {
+				status.Success(fmt.Sprintf("Service %s already exported", svc.Name))
+				continue
+			}
+
+			return status.Error(err, "Failed to export Service")
+		}
+
+		status.Success(fmt.Sprintf("Service %s exported successfully", svc.Name))
```
skitt: This could re-use `Export()` instead of effectively duplicating it — please change `Export()` to include the name of the service in its status messages, and then change `Exports()` to call `Export()`.

skitt (changes requested):
This is great, thanks! It would be even better if `Exports()` called `Export()`, see my review comment for details.

Also, could you add a system test for this? Look at `export_service()` in https://github.com/submariner-io/subctl/blob/devel/scripts/test/system.sh#L58

pkg/service/export.go
Xunzhuo: This is a bit different. `Exports()` needs to list all svcs under the namespace. Maybe I can reuse Export in `for-range`

pkg/service/export.go
```diff
@@ -54,14 +55,34 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
...
+			status.Success(fmt.Sprintf("Service %s/%s already exported", serviceNamespace, svcName))
 			return nil
 		}
 
-		return status.Error(err, "Failed to export Service")
+		return status.Error(err, fmt.Sprintf("Failed to export Service %s/%s", serviceNamespace, svcName))
 	}
 
-	status.Success("Service exported successfully")
+	status.Success(fmt.Sprintf("Service %s/%s exported successfully", serviceNamespace, svcName))
```
skitt, line 65:
```suggestion
	status.Success("Service %s/%s exported successfully", serviceNamespace, svcName)
```

pkg/service/export.go
```diff
@@ -54,14 +55,34 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
 		Create(context.TODO(), resourceServiceExport, metav1.CreateOptions{})
 	if err != nil {
 		if k8serrors.IsAlreadyExists(err) {
-			status.Success("Service already exported")
+			status.Success(fmt.Sprintf("Service %s/%s already exported", serviceNamespace, svcName))
```
skitt, line 58:
```suggestion
			status.Success("Service %s/%s already exported", serviceNamespace, svcName)
```

pkg/service/export.go
```diff
@@ -54,14 +55,34 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
...
 			return nil
 		}
 
-		return status.Error(err, "Failed to export Service")
+		return status.Error(err, fmt.Sprintf("Failed to export Service %s/%s", serviceNamespace, svcName))
```
skitt, line 62:
```suggestion
		return status.Error(err, "Failed to export Service %s/%s", serviceNamespace, svcName)
```

pkg/service/export.go
```diff
@@ -54,14 +55,34 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
...
+		return status.Error(err, "Unable to list the Services in namespace %q", serviceNamespace)
+	}
+
+	if len(svcs.Items) == 0 {
+		status.Warning("No Services exist in target namespace")
+		return nil
+	}
+
+	for _, svc := range svcs.Items {
+		if err := Export(clientProducer, svc.Namespace, svc.Name, status); err != nil {
```
skitt, line 82: I wonder what behaviour would be more useful:

* stop at the first error (as happens now)
* try to export all services in the namespace, accumulating errors

What do you think?

skitt (changes requested):
Thanks, a few minor fixes and an open question (the latter non-blocking IMO).

pkg/service/export.go
skitt: Yes, my comment applied to the highlighted code only, not the `for` loop itself — I am well aware that `Exports()` and `Export()` are different :wink:.

cmd/subctl/export.go
```diff
@@ -54,6 +54,21 @@ var (
 				}, cli.NewReporter()))
 		},
 	}
+	exportNamespaceCmd = &cobra.Command{
+		Use:   "namespace <namespaceName>",
```
tpantelis, line 58: This indicates to me that the command exports an actual namespace resource. How about we name it “services”?

Xunzhuo: Like 

subctl export services -n xxxx 

?

skitt: Right, I wondered about that but hadn’t thought of a decent alternative — the risk with `export namespace` is that users will think that any changes made to the namespace afterwards will automatically be reflected.

Xunzhuo: One more thing, I am still thinking, if we need to export the services created by submariner itself when exporting services under a specific namespace?

Xunzhuo: If people should be aware of that, by adding a new flag when using `export namespace` like `--include-submariner-services`?

pkg/service/export.go
```diff
@@ -54,14 +55,34 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
...
+		return status.Error(err, fmt.Sprintf("Failed to export Service %s/%s", serviceNamespace, svcName))
 	}
 
-	status.Success("Service exported successfully")
+	status.Success(fmt.Sprintf("Service %s/%s exported successfully", serviceNamespace, svcName))
+
+	return nil
+}
+
+func Exports(clientProducer client.Producer, serviceNamespace string, status reporter.Interface) error {
```
tpantelis, line 70:
```suggestion
func ExportAll(clientProducer client.Producer, inNamespace string, status reporter.Interface) error {
```

cmd/subctl/export.go
skitt: Which services are you referring to? Submariner doesn’t create services apart from the metrics services in `submariner-operator`.

tpantelis: FWIR, we do create internal services for GN as well.  If we have to add complexity to exclude certain services, that kind of throws up a red flag for me as to whether we should provide this command. I’m not sure how common it would be for a user to blindly export all services in a namespace. If they really want to do that, they could easily write a simple script to do it. Also this would be a one-time export which would diminish its usefulness if a user expects it to be dynamic.

skitt: Ah yes, the externalIP services...

sridhargaddam: Agree with @tpantelis. Submariner creates (and owns) an internal service for every exported service in the same namespace where the original service resides. Even I feel that the requirement can be easily achieved with a simple script.

skitt:
In its current state this would end up confusing users. It might be useful to have a general “export all services in a namespace” feature, but it would at least have to handle Globalnet services, and ideally would involve a new MCS object and periodic synchronization to export services added to a namespace.

PR #913, opened 2023-09-13
Use endpoints from Broker while validating overlapping CIDRs

sridhargaddam:
As part of the following PR[1], we no longer sync an endpoint from the Broker if the endpoint has any overlapping CIDRs with the local cluster. This was done to avoid issues on the Gateway node. Consequently, the subctl diagnose code should ideally examine the endpoints on Broker when validating overlapping CIDRs instead of inspecting the endpoints on the local cluster. This PR addresses this issue.

[1] https://github.com/submariner-io/submariner/pull/2263

### DIFF
--- a/pkg/diagnose/deployments.go
+++ b/pkg/diagnose/deployments.go
@@ -26,6 +26,8 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/subctl/internal/constants"
+	"github.com/submariner-io/subctl/internal/restconfig"
+	"github.com/submariner-io/subctl/pkg/client"
 	"github.com/submariner-io/subctl/pkg/cluster"
 	submarinerv1 "github.com/submariner-io/submariner/pkg/apis/submariner.io/v1"
 	"github.com/submariner-io/submariner/pkg/cidr"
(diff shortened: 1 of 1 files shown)


pkg/diagnose/deployments.go
```diff
@@ -68,10 +70,20 @@ func checkOverlappingCIDRs(clusterInfo *cluster.Info, status reporter.Interface)
...
+	}
+
 	endpointList := &submarinerv1.EndpointList{}
 
-	err := clusterInfo.ClientProducer.ForGeneral().List(context.TODO(), endpointList,
-		controllerClient.InNamespace(clusterInfo.Submariner.Namespace))
+	err = clientProducer.ForGeneral().List(context.TODO(), endpointList,
+		controllerClient.InNamespace(brokerNamespace))
 	if err != nil {
 		return status.Error(err, "Error listing the Submariner endpoints")
```
yboaron: Overall code LGTM, I have one comment/Q,
What if the broker is down? maybe we can update the error message?

sridhargaddam: If there is an error while connecting to the Broker, the `err` message should have the appropriate details.

sridhargaddam: Let me add some note saying that its while reading from the Broker.

yboaron (approved):

skitt (approved):

vthapar (approved):

PR #966, opened 2023-10-06
Use generic resource handling

skitt:
Depends on https://github.com/submariner-io/admiral/pull/756
Depends on https://github.com/submariner-io/admiral/pull/760
Depends on https://github.com/submariner-io/cloud-prepare/pull/817
Depends on https://github.com/submariner-io/lighthouse/pull/1402
Depends on https://github.com/submariner-io/submariner/pull/2742
Depends on https://github.com/submariner-io/submariner-operator/pull/2839


pkg/serviceaccount/ensure.go
```diff
@@ -45,8 +44,8 @@ const (
 
 func ensure(ctx context.Context, kubeClient kubernetes.Interface, namespace string, sa *corev1.ServiceAccount) (bool, error) {
 	result, err := util.CreateOrUpdate(ctx, resource.ForServiceAccount(kubeClient, namespace), sa,
```
tpantelis, line 46:
```suggestion
	result, err := util.CreateOrUpdate[*corev1.ServiceAccount](ctx, resource.ForServiceAccount(kubeClient, namespace), sa,
```

skitt: I rather like letting the compiler infer the type :wink:

tpantelis: yeah but you've specified the type explicitly in other places so for consistency.... but it doesn't matter either way, that's why I didn't -1 😄

skitt: Ah, I thought I’d dropped the type parameters everywhere they weren’t needed, but I missed some.

tpantelis (approved):

sridhargaddam (approved):

PR #981, opened 2023-10-25
Reduce string constant duplication

tpantelis:
Newer versions of `goconst` detect more instances of string duplication so refactor the offending strings to avoid an error.

### DIFF
--- a/cmd/subctl/aws.go
+++ b/cmd/subctl/aws.go
@@ -53,8 +53,8 @@ var (
 	awsCleanupCmd = &cobra.Command{
 		Use:   "aws",
 		Short: "Clean up an AWS cloud",
-		Long: "This command cleans up an OpenShift installer-provisioned infrastructure (IPI) on AWS-based" +
-			" cloud after Submariner uninstallation.",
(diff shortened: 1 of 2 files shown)


cmd/subctl/aws.go
```diff
@@ -53,8 +53,8 @@ var (
 	awsCleanupCmd = &cobra.Command{
 		Use:   "aws",
 		Short: "Clean up an AWS cloud",
-		Long: "This command cleans up an OpenShift installer-provisioned infrastructure (IPI) on AWS-based" +
-			" cloud after Submariner uninstallation.",
+		Long: "This command cleans up an OpenShift installer-provisioned infrastructure (IPI) on " +
```
sridhargaddam, line 56: Am I missing something. Where is the duplication reduced in this PR?

tpantelis: The partial text " _cloud after Submariner uninstallation._" for wrapping. I just refactored the wrapped part so it no longer appears as a dupe. Smoke and mirrors 😄

sridhargaddam (approved):

skitt (approved):

PR #1016, opened 2023-11-30
diagnose: handle undetermined network plugins

### LINKED ISSUE #479: subctl diagnose gives incorrect information when the CNI is unknown
**What happened**:

Running `subctl diagnose all` during Submariner startup, before the CNI has been identified produces

```
 ✗ Checking Submariner support for the CNI network plugin
 ✗ The detected CNI plugin ("") is not supported by Submariner. Supported plugins: [generic canal-flannel weave-net OpenShiftSDN OVNKubernetes calico kindnet]
```

`subctl` should indicate that the CNI hasn’t been determined yet or is unknown, not that the CNI is not supported.

Set up Submariner, run `subctl diagnose all` while Submariner is starting up.

**Environment**:
- Cloud provider or hardware configuration: AWS
- Install tools: ACM 2.7
- Others:

skitt:
When the network plugin hasn't been determined yet, instead of indicating that it's not supported, explicitly warn that it hasn't been determined yet.

Fixes: #479


pkg/diagnose/cni.go
```diff
@@ -62,6 +62,11 @@ func CNIConfig(clusterInfo *cluster.Info, _ string, status reporter.Interface) e
 	status.Start("Checking Submariner support for the CNI network plugin")
 	defer status.End()
 
+	if clusterInfo.Submariner.Status.NetworkPlugin == "" {
+		status.Warning("The network plugin hasn't been determined yet.")
```
tpantelis, line 66: This implies it would be determined at some point? But shouldn't L85 below handle this case?

skitt: Yes, it is determined later. Line 85 handles the case but produces a confusing message, indicating that the plugin isn’t supported when the reality is that Submariner doesn’t know what the plugin is yet.

tpantelis: That's L80. L8(5|6) says "Submariner could not detect the CNI network plugin and is using (%q) plugin.". Isn't that the message we want? Perhaps we just need to move L85-88 here?

skitt: Oops, yes. But that message would be inaccurate: if `NetworkPlugin` is empty in the status, Submariner isn’t using the generic plugin, it’s just not detected the plugin yet.

tpantelis: Ahh right - this is for diagnose.

tpantelis (approved):

pkg/diagnose/cni.go
sridhargaddam: I feel the message `The network plugin hasn't been determined yet` could give an impression to the user that if they retry the diagnose command after a while, the CNI plugin could be determined by then.

dfarrell07 (approved):

pkg/diagnose/cni.go
skitt: @sridhargaddam yes, that’s exactly what should happen. The situation handled here only happens if `subctl diagnose` is run before the plugin has been identified; once it has, `diagnose` will show the type (or “generic” if it isn’t known).

sridhargaddam: got it, thanks for the clarification.

PR #1048, opened 2024-01-15
Add --extracontext arg to subctl verify

### LINKED ISSUE #276: Add a three-context variant for subctl verify

`subctl verify` should support three contexts.

**Why is this needed**:

Some Lighthouse tests only run if three contexts are available; it would be good to be able to run them using `subctl verify`.

tpantelis:
...to run lighthouse tests that require a third cluster.

Added a system test using the new arg that is run when **lighthouse** is deployed. This requires the cluster settings to specify a third cluster which can be problematic wrt CI env resources. So the System Test **globalnet** and **lighthouse** options were split into 2 jobs with **lighthouse** using a separate cluster settings containing the third cluster and only the control-plane node.

Fixes https://github.com/submariner-io/subctl/issues/276


cmd/subctl/verify.go
```diff
@@ -62,18 +62,21 @@ var (
...
 	Short: "Run verifications between two clusters",
-	Long: `This command performs various tests to verify that a Submariner deployment between two clusters
-is functioning properly. The verifications performed are controlled by the --only and --enable-disruptive
-flags. All verifications listed in --only are performed with special handling for those deemed as disruptive.
-A disruptive verification is one that changes the state of the clusters as a side effect. If running the
-command interactively, you will be prompted for confirmation to perform disruptive verifications unless
-the --enable-disruptive flag is also specified. If running non-interactively (that is with no stdin),
---enable-disruptive must be specified otherwise disruptive verifications are skipped.
+	Long: `This command performs various tests to verify that a Submariner deployment between two clusters,
+specified via the --context and --tocontext args, is functioning properly. Some tests require a third cluster,
```
vthapar: How about specifying "Some servicediscovery tests require a third cluster" to make it clear that if someone is not using servicediscovery verify they don't need to use this argument.

Or, if service discovery is not enabled for verification give a warn or info message to user that `--extracontext` will be ignored.

vthapar (changes requested):

cmd/subctl/verify.go
tpantelis: I was thinking not to make it specific to service discovery in case we ever add other test(s) that need a third cluster.

aswinsuryan: Making it genric may be a good idea as there could be use cases later for newer test in connectivity requiring a third cluster.

aswinsuryan (approved):

cmd/subctl/verify.go
vthapar: > I was thinking not to make it specific to service discovery in case we ever add other test(s) that need a third cluster.

We can change the info message when we need it for other suites. As of now it is only service discovery, will help user know that this will apply only to service discovery and other tests don't need this option.

tpantelis: yeah I agree it would be useful. We can change it later if needed although that kind of thing tends to slip through the cracks.

tpantelis: Done

skitt (approved):

vthapar (approved):

PR #1086, opened 2024-03-27
Enhance --image-override verification

### LINKED ISSUE #601: --image-override doesn’t always check the components
**What happened**:

On commands such as `subctl diagnose`, specifying an `--image-override` with an invalid component doesn’t produce an error.

`subctl join` with an invalid component produces an error; I expected other commands supporting `--image-override` to perform the same check.

`subctl diagnose --image-override=nettest=foo`

(the correct component is `submariner-nettest`).

The check is handled in `MergeImageOverrides`.

tpantelis:
The commands that support the `--image-override` flag should verify it up front via the `Arg` function since its user input rather than during the command execution so it doesn't appear as a failure.

Also consolidate the flag setup so each command uses a common `imageOverrides` variable instead of commands defining their own variable.

Fixes https://github.com/submariner-io/subctl/issues/601


scripts/test/system.sh
```diff
@@ -192,6 +192,11 @@ _subctl diagnose firewall nat-discovery --validation-timeout 20 --context cluste
 _subctl diagnose firewall inter-cluster --validation-timeout 20 "${KUBECONFIGS_DIR}"/kind-config-cluster1 "${KUBECONFIGS_DIR}"/kind-config-cluster2 && exit 1
 # Obsolete firewall nat-discovery variant
 _subctl diagnose firewall nat-discovery --validation-timeout 20 "${KUBECONFIGS_DIR}"/kind-config-cluster1 "${KUBECONFIGS_DIR}"/kind-config-cluster2 && exit 1
+# Invalid image-override flag
+_subctl diagnose all --image-override=foo=bar && exit 1
+_subctl diagnose firewall inter-cluster -kubeconfig "${KUBECONFIGS_DIR}"/kind-config-cluster1 --remoteconfig "${KUBECONFIGS_DIR}"/kind-config-cluster2 --image-override=foo=bar && exit 1
```
skitt:
```suggestion
_subctl diagnose firewall inter-cluster --kubeconfig "${KUBECONFIGS_DIR}"/kind-config-cluster1 --remoteconfig "${KUBECONFIGS_DIR}"/kind-config-cluster2 --image-override=foo=bar && exit 1
```

skitt (changes requested):

vthapar (approved):

skitt (approved):

PR #1103, opened 2024-04-10
Add HTTP proxy args to subctl join, deploy-broker and upgrade commands

tpantelis:
...and propagate to the operator pod env.

See discussion [here](https://github.com/submariner-io/submariner-operator/pull/3013#issuecomment-2035366304) for more context.

### DIFF
--- a/pkg/join/options.go
+++ b/pkg/join/options.go
@@ -18,6 +18,8 @@ limitations under the License.
 
 package join
 
+import "golang.org/x/net/http/httpproxy"
+
 type Options struct {
 	PreferredServer               bool
 	ForceUDPEncaps                bool
@@ -46,4 +48,5 @@ type Options struct {
 	BrokerURL                     string
 	CustomDomains                 []string
 	ImageOverrideArr              []string
+	HTTPProxyConfig               httpproxy.Config
 }
--- a/pkg/operator/deployment/ensure.go
+++ b/pkg/operator/deployment/ensure.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/names"
 	"github.com/submariner-io/subctl/pkg/deployment"
+	"golang.org/x/net/http/httpproxy"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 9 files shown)


pkg/operator/deployment/ensure.go
```diff
@@ -122,3 +123,20 @@ func GetPodLabelSelector(kubeClient kubernetes.Interface, namespace string) (str
...
+	return vars
+}
+
+func appendEnvVarIfValue(vars []v1.EnvVar, name, value string) []v1.EnvVar {
+	if value != "" {
+		vars = append(vars, v1.EnvVar{Name: name, Value: value})
+	}
+
+	return vars
+}
```
tpantelis: I can resurrect https://github.com/submariner-io/submariner-operator/pull/3016 if folks approve internally propagating the settings via the local env instead of passing via function args.

pkg/join/options.go
```diff
@@ -44,6 +44,9 @@ type Options struct {
 	CableDriver                   string
 	CoreDNSCustomConfigMap        string
 	BrokerURL                     string
+	HTTPProxy                     string
+	HTTPSProxy                    string
+	HTTPNoProxy                   string
```
skitt: How about

```suggestion
	NoProxy                   string
```

to match `httpproxy.Config`?

pkg/operator/deployment/ensure.go
```diff
@@ -122,3 +123,20 @@ func GetPodLabelSelector(kubeClient kubernetes.Interface, namespace string) (str
 
 	return labels.SelectorFromSet(dep.Spec.Template.ObjectMeta.Labels).String(), nil
 }
+
+func addHTTPProxyEnvVars(vars []v1.EnvVar) []v1.EnvVar {
+	proxyEnv := httpproxy.FromEnvironment()
```
skitt: I would rather have an explicit instance of the options passed in here, rather than using the environment as global variables. Case in point, since this only touches `join`, the end result is that operator deployments from `deploy-broker` and `upgrade` will be silently affected by any local proxy settings.

Relying on environment variables makes it too easy to produce a broken setup without the user knowing why; even if the existing commands are fixed to behave properly, this adds a gotcha to any future sub-command involving an operator deployment.

skitt (changes requested):
It would be good if this also handled `deploy-broker` and `upgrade`, since the operators deployed there might need proxy settings too.

pkg/operator/deployment/ensure.go
tpantelis: yeah good points - I was a little wary of passing via env vars.

skitt (approved):

vthapar (approved):

PR #1128, opened 2024-05-15
Set the operator up with enforced secrets

skitt:
To prevent arbitrary secret access using compromised SAs, the SAs created by Submariner are now configured to enforce mountable secrets. This requires that accessible secrets be listed explicitly in the SA. To make this simple, use a static name for the broker secret. To allow secrets to be configured, leave them alone when creating or updating the SAs.

Depends on https://github.com/submariner-io/submariner-operator/pull/3064
Depends on https://github.com/submariner-io/submariner-operator/pull/3067

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/submariner-io/lighthouse v0.18.0-m3
 	github.com/submariner-io/shipyard v0.18.0-m3
 	github.com/submariner-io/submariner v0.18.0-m3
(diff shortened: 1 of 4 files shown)


dfarrell07 (approved):

aswinsuryan (approved):

tpantelis (approved):

dfarrell07:
We were hoping to see this work in 0.18, including d/s, before backporting it.

PR #1131, opened 2024-05-19
Diagnose deployment - verify that only single GW pod labeled as active

yboaron:
We recently ran into an issue where some tests in subctl verify failed because more than a single GW pod was labeled as active.

This PR extends 'subctl diagnose deployment' to ensure that only a single GW pod is labeled as active.

### DIFF
--- a/pkg/diagnose/deployments.go
+++ b/pkg/diagnose/deployments.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/names"
 	"github.com/submariner-io/admiral/pkg/reporter"
 	"github.com/submariner-io/subctl/internal/constants"
 	"github.com/submariner-io/subctl/internal/restconfig"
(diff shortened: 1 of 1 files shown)


yboaron:
cc @aswinsuryan @tpantelis

pkg/diagnose/deployments.go
```diff
@@ -200,8 +201,15 @@ func checkPodsStatus(k8sClient kubernetes.Interface, namespace string, status re
 		return
 	}
 
+	numOfActiveGws := 0
+
 	for i := range pods.Items {
 		pod := &pods.Items[i]
+
+		if (pod.Labels["app"] == names.GatewayComponent) && (pod.Labels["gateway.submariner.io/status"] == "active") {
```
tpantelis: We could the status constant here

```suggestion
		if (pod.Labels["app"] == names.GatewayComponent) && (pod.Labels["gateway.submariner.io/status"] == submarinerv1.HAStatusActive) {
```

pkg/diagnose/deployments.go
```diff
@@ -214,4 +222,8 @@ func checkPodsStatus(k8sClient kubernetes.Interface, namespace string, status re
 			}
 		}
 	}
+
+	if numOfActiveGws != 1 {
+		status.Warning("Number of GW pods labeled as active is not 1, (numOfActiveGws is %d)", numOfActiveGws)
```
tpantelis:
```suggestion
		status.Warning("Expected one Gateway pod to be labeled as active. Found %d", numOfActiveGws)
```

I think it may also be useful to identify which gateways were labeled as active. Perhaps the pod's node  name would be most useful to report. We can also use the `gateway.submariner.io/node` label. So instead of `numOfActiveGws`, we could define an `activeGwNodeNames` array and check the len here and report `status.Warning("... Found %d on nodes %v", len(activeGwNameNames), activeGwNameNames)`.

tpantelis (approved):

tpantelis (approved):

dfarrell07 (approved):

PR #1147, opened 2024-06-04
Drop the dependency on wireguard in e2e

skitt:
The dependency breaks Windows builds.


tpantelis:
The cross-build job doesn't run on draft PRs.

vthapar (approved):

tpantelis (approved):

tpantelis:
Shouldn't we merge this?

tpantelis:
I updated the commit hash.

skitt:
> Shouldn't we merge this?

Current `devel` isn’t affected by this, so there’s no urgency. The dependencies will all be aligned when v0.19.0-m0 is released.

tpantelis:
> > Shouldn't we merge this?
> 
> Current `devel` isn’t affected by this, so there’s no urgency. The dependencies will all be aligned when v0.19.0-m0 is released.

yes but that won't be for a while and someone may push a subctl PR before that.

skitt:
> > > Shouldn't we merge this?
> > 
> > 
> > Current `devel` isn’t affected by this, so there’s no urgency. The dependencies will all be aligned when v0.19.0-m0 is released.
> 
> yes but that won't be for a while and someone may push a subctl PR before that.

Current `devel` isn’t affected by the dependency change.

PR #1195, opened 2024-07-31
Improve version check

vthapar:
Improves version check by making following changes:

1. Look for version in Submariner.Status and ServiceDiscovery.Spec. These two are populated by submariner-oprator so are always present and accurate. This also fixes a bug in current code where if Submariner.Spec.Version is missing, it doesn't check for it in Status or ServiceDiscovery.

2. For subctl commands the major and minor versions are more important than patch versions. This is because changes that significantly affect subctl commands are mostly done in releases with subsequent releases being mostly bug fixes.

### DIFF
--- a/internal/restconfig/restconfig.go
+++ b/internal/restconfig/restconfig.go
@@ -205,10 +205,10 @@ func (rcp *Producer) RunOnSelectedContext(function PerContextFn, status reporter
 	}
 
 	var submVersion string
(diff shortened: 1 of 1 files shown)


vthapar:
@maayanf24 Draft of the changes I had. Needs more testing, but should work. Would be good to discuss the design changes with team, or in an EP.

PR #1228, opened 2024-09-16
Check route agent connections in subctl diagnose

tpantelis:
Depends on https://github.com/submariner-io/submariner/pull/3166


dfarrell07 (approved):

pkg/diagnose/connections.go
```diff
@@ -76,7 +79,47 @@ func Connections(clusterInfo *cluster.Info, _ string, status reporter.Interface)
...
+func checkRouteAgentConnections(clusterInfo *cluster.Info, status reporter.Interface) error {
+	status.Start("Checking route agent connections")
+	defer status.End()
+
+	routeAgents, err := clusterInfo.GetRouteAgents()
+	if err != nil {
+		return status.Error(err, "Error retrieving route agents")
+	}
+
+	if len(routeAgents) == 0 {
```
yboaron, line 97: Do we need to consider if submariner.healthcheck != enabled  ?

yboaron (approved):

pkg/diagnose/connections.go
tpantelis: We create a `RouteAgent` resource regardless (same with `Gateways`).

aswinsuryan (approved):

PR #1229, opened 2024-09-20
Use cobra's version flag support

skitt:
Assigning a value to the Version field in the root cobra command automatically defines a --version flag which displays the command name and version, in much the same way as "subctl version" does already. This provides version information in a fashion aligned with common practice.


dfarrell07 (approved):

tpantelis:
Should we remove our version command?

vthapar (approved):

vthapar:
> Should we remove our version command?

There might be automated scripts using it. We should deprecate it before removing it.

PR #1231, opened 2024-09-24
Add support for clustersetIP

vthapar:
1. Add ClustersetIP flags to deploy-broker and join.
2. Add ClustersetIP CIDRs to config CRs.
3. Show clustersetupCIDR in `show networks` output.
4. Add use-clusterset-ip flag to `service export`

Refer: https://github.com/submariner-io/enhancements/pull/230


cmd/subctl/deploybroker.go
```diff
@@ -89,6 +90,10 @@ func addDeployBrokerFlags(flags *pflag.FlagSet) {
 
 	flags.StringVar(&deployflags.BrokerURL, "broker-url", "",
 		"broker API endpoint URL (stored in the broker information file, defaults to the context URL)")
+	flags.BoolVar(&deployflags.BrokerSpec.ClustersetIPEnabled, "enable-clusterset-ip", false,
+		"set default support for ClustersetIP in connecting clusters (default disabled)")
```
tpantelis:
```suggestion
		"set default support for use of clusterset IP for exported services in connecting clusters (default disabled)")
```

cmd/subctl/deploybroker.go
```diff
@@ -89,6 +90,10 @@ func addDeployBrokerFlags(flags *pflag.FlagSet) {
...
+	flags.StringVar(&deployflags.BrokerSpec.ClustersetIPCIDRRange, "clusterset-ip-cidr-range",
+		clustersetip.DefaultCIDR, "ClustersetIP CIDR supernet range for allocating ClustersetIP CIDRs to each cluster")
```
tpantelis:
```suggestion
		clustersetip.DefaultCIDR, "Clusterset IP CIDR supernet range for allocating clusterset IP CIDRs to each cluster")
```

cmd/subctl/export.go
```diff
@@ -72,3 +73,8 @@ func validateArguments(args []string) error {
 
 	return nil
 }
+
+func addExportServiceFlags(cmd *cobra.Command) {
+	exportRestConfigProducer.SetupFlags(exportServiceCmd.Flags())
+	cmd.PersistentFlags().BoolVar(&useClustersetIP, "use-clusterset-ip", false, "use clusterset ip for this service")
```
tpantelis:
```suggestion
	cmd.PersistentFlags().BoolVar(&useClustersetIP, "use-clusterset-ip", false, "use clusterset IP for this service")
```

cmd/subctl/join.go
```diff
@@ -147,6 +147,10 @@ func addJoinFlags(cmd *cobra.Command) {
 		"check the broker certificate (disable this to allow \"insecure\" connections)")
 	cmd.Flags().StringVar(&joinFlags.BrokerURL, "broker-url", "",
 		"URL of the broker API endpoint (overrides the URL stored in the broker information file)")
+	cmd.Flags().BoolVar(&joinFlags.EnableClustersetIP, "enable-clusterset-ip", false,
+		"set default support for ClustersetIP in the cluster (default disabled)")
```
tpantelis:
```suggestion
		"set default support for use of clusterset IP for exported services in the cluster (default disabled)")
```

cmd/subctl/join.go
```diff
@@ -147,6 +147,10 @@ func addJoinFlags(cmd *cobra.Command) {
...
+	cmd.Flags().StringVar(&joinFlags.ClustersetIPCIDR, "clusterset-ip-cidr", "",
+		"ClustersetIP CIDR to be allocated to the cluster")
```
tpantelis:
```suggestion
		"Clusterset IP CIDR to be allocated to the cluster")
```

pkg/deploy/submariner.go
```diff
@@ -46,6 +47,7 @@ type SubmarinerOptions struct {
 	LoadBalancerEnabled           bool
 	HealthCheckEnabled            bool
 	BrokerK8sInsecure             bool
+	ClustersetipEnabled           bool
```
tpantelis:
```suggestion
	ClustersetIPEnabled           bool
```

pkg/deploy/submariner.go
```diff
@@ -105,6 +108,7 @@ func populateSubmarinerSpec(options *SubmarinerOptions, brokerInfo *broker.Info,
 		BrokerK8sApiServer:       brokerURL,
 		BrokerK8sSecret:          brokerSecret.ObjectMeta.Name,
 		BrokerK8sInsecure:        options.BrokerK8sInsecure,
+		ClustersetIPEnabled:      options.ClustersetipEnabled,
```
tpantelis:
```suggestion
		ClustersetIPEnabled:      options.ClustersetipEnabled,
		ClustersetIPCIDR:           clustersetConfig.ClustersetIPCIDR,
```

pkg/deploy/submariner.go
```diff
@@ -126,6 +130,10 @@ func populateSubmarinerSpec(options *SubmarinerOptions, brokerInfo *broker.Info,
 		submarinerSpec.GlobalCIDR = netconfig.GlobalCIDR
 	}
 
+	if clustersetConfig.ClustersetIPCIDR != "" {
+		submarinerSpec.ClustersetIPCIDR = clustersetConfig.ClustersetIPCIDR
+	}
```
tpantelis: We can just initialize above regardless if empty.

```suggestion
```

pkg/join/join.go
```diff
@@ -98,6 +104,12 @@ func ClusterToBroker(ctx context.Context, brokerInfo *broker.Info, options *Opti
 		}
 	}
 
+	err = clustersetip.AllocateCIDRFromConfigMap(ctx, brokerClientProducer.ForGeneral(), brokerNamespace,
+		&clustersetConfig, status)
+	if err != nil {
+		return errors.Wrap(err, "unable to determine clustersetIP CIDR")
```
tpantelis:
```suggestion
		return errors.Wrap(err, "unable to determine the clusterset IP CIDR")
```

pkg/service/export.go
```diff
@@ -43,6 +44,10 @@ func Export(clientProducer client.Producer, serviceNamespace, svcName string, st
 		},
 	}
 
+	if useClustersetIP {
+		mcsServiceExport.SetAnnotations(map[string]string{lhconstants.UseClustersetIP: "true"})
```
tpantelis:
```suggestion
		mcsServiceExport.SetAnnotations(map[string]string{lhconstants.UseClustersetIP: strconv.FormatBool(true)})
```

tpantelis (changes requested):

pkg/deploy/broker.go
```diff
@@ -82,6 +88,15 @@ func Broker(options *BrokerOptions, clientProducer client.Producer, status repor
...
 	}
 
+	if err = clustersetip.ValidateExistingClustersetIPNetworks(ctx, clientProducer.ForGeneral(), options.BrokerNamespace); err != nil {
+		return status.Error(err, "error validating existing clustersetIPCIDR configmap")
+	}
+
+	if err = clustersetip.CreateConfigMap(ctx, clientProducer.ForGeneral(), options.BrokerSpec.ClustersetIPEnabled,
+		options.BrokerSpec.ClustersetIPCIDRRange, 0, options.BrokerNamespace); err != nil {
+		return status.Error(err, "error creating clustersetIPCIDR configmap on Broker")
+	}
```
tpantelis: Shouldn't the [broker controller](https://github.com/submariner-io/submariner-operator/blob/268a9960c6b70ddbf6fd751fe6eae6ebb24f4077/controllers/submariner/broker_controller.go#L85) handle this as it does for globalnet?

vthapar: > Shouldn't the [broker controller](https://github.com/submariner-io/submariner-operator/blob/268a9960c6b70ddbf6fd751fe6eae6ebb24f4077/controllers/submariner/broker_controller.go#L85) handle this as it does for globalnet?

Yeah, it should but that is for addon use case where addon operator runs broker controller. There is no broker controller when usin subctl to deploy broker, that is why we do this for globlanet too in subctl.

https://github.com/submariner-io/subctl/blob/e6b9fe5154cca34f9fe29b7dc7ee679bfbd8da23/pkg/deploy/broker.go#L80

pkg/service/export.go
vthapar: This is the tricky one. Way flags work, default is false. But I don't want to set anything if user hasn't given an explicit option. In this case, setting it to `false` is different from not setting it at all. Not setting it means it will use the global setting on cluster. Setting false means even if global is true, it will not use clustersetip for this service.

I realize the way I've done means user can never set `false` explicitly, but I went with my option for backwards compatibility where existing usage is to not have it set at all.

Any suggestions on how we can handle it better? This is a case where we want true, false, nil. Can we do that?

EDIT: How about make it a string?

pkg/deploy/submariner.go
vthapar: Ah right, forgot that this is `omitifempty`.

cmd/subctl/deploybroker.go
vthapar: done

cmd/subctl/deploybroker.go
vthapar: done

cmd/subctl/export.go
vthapar: done

cmd/subctl/join.go
vthapar: done

cmd/subctl/join.go
vthapar: done

pkg/deploy/submariner.go
vthapar: Done

pkg/deploy/submariner.go
vthapar: Done

pkg/deploy/broker.go
tpantelis: Actually, the broker controller always runs in the operator pod - it's activated by creation of the `Broker` resource.

The behavior wrt the addon vs subctl is actually the opposite: the addon creates the globalnet `ConfigMap` and does not create the `Broker` resource while subctl does both tho I'm not sure why.  It's fine if we do the same for clusterset IP here to mirror globalnet behavior but perhaps we should also do the same in the broker controller otherwise it renders the new fields in the Broker CRD meaningless. We could revisit  the duplicate behavior later. 

Also note that the helm charts does neither but it's moot b/c we expect the user to pass in the global CIDR when deploying the operator.

pkg/service/export.go
tpantelis: yeah make it a string and verify it's either "true" or "false".

tpantelis (approved):

PR #1242, opened 2024-10-02
Bump lighthouse to fix e2e

vthapar:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/submariner-io/admiral v0.19.0-rc0
 	github.com/submariner-io/cloud-prepare v0.19.0-rc0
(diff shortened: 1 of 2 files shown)


skitt:
Heh, now we have a commit saying “Do Not Merge this” in the history ;-).

tpantelis:
> Heh, now we have a commit saying “Do Not Merge this” in the history ;-).

Crap. Sorry I forgot to update the commit too in my haste.

skitt:
> Crap. Sorry I forgot to update the commit too in my haste.

Wasn’t the commit redundant anyway? The dependency will be bumped by the release process.

tpantelis:
> > Crap. Sorry I forgot to update the commit too in my haste.
> 
> Wasn’t the commit redundant anyway? The dependency will be bumped by the release process.

Yes but the release process runs E2E which uses subctl which happens before the bump.

PR #1282, opened 2025-01-02
Temporarily disable goheader on pkg/join/options.go

skitt:
This triggers a false positive in golangci-lint 1.63.2.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -207,3 +207,9 @@ issues:
     - path: cmd/subctl/.*\.go
       linters:
         - goheader
+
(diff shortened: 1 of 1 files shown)


tpantelis:
Hmm... why the false positive just on that one file...

tpantelis:
For some reason, `subctl diagnose` is failing when checking the metrics availability but only when globalnet is enabled.

tpantelis (approved):

tpantelis:
Included with https://github.com/submariner-io/subctl/pull/1284

PR #1396, opened 2025-05-16
Migrate to golangci-lint v2

skitt:
This also fixes a few issues:
* simplify kubeproxy diagnose using De Morgan's laws
* fix broker Info JSON tags
* name the Logger V() argument

musttag is left disabled because it requires fixes in Azure auth which must match the current file format.

Depends on https://github.com/submariner-io/shipyard/pull/1963

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,63 +1,86 @@
 ---
-linters-settings:
-  depguard:
-    rules:
-      main:
-        deny:
-          - pkg: github.com/mcuadros/go-version
-            desc: Use coreos/go-semver instead
-  gocritic:
-    enabled-tags:
-      - diagnostic
-      - opinionated
-      - performance
-      - style
-    disabled-checks:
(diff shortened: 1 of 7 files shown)


dfarrell07:
[Subctl Commands (lighthouse)](https://github.com/submariner-io/subctl/actions/runs/15116270805/job/42487624285?pr=1396#logs) failed, re-running (didn't look ephemeral though).

dfarrell07 (approved):

vthapar (approved):

vthapar:
Failing test will be fixed by https://github.com/submariner-io/subctl/pull/1400 So now we need to force merge at least one of these to get the other to pass.

skitt:
> Failing test will be fixed by #1400 So now we need to force merge at least one of these to get the other to pass.

Or we can combine the two, which is what I just did here (cherry-picking your commit).

PR #1472, opened 2025-08-07
Use non-root user in Konflux Dockerfile

dfarrell07:
Fixes an openshift-preflight check.

### DIFF
--- a/package/Dockerfile.subctl.konflux
+++ b/package/Dockerfile.subctl.konflux
@@ -31,6 +31,8 @@ ARG BASE_BRANCH
 ARG SOURCE
 ARG TARGETPLATFORM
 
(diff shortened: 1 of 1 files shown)


dfarrell07:
This fixes `ecosystem-cert-preflight-checks`, but there's a new error that seems to have started without a clear cause.

```
                "name": "RunAsNonRoot",
                "elapsed_time": 0,
                "description": "Checking if container runs as the root user because a container that does not specify a non-root user will fail the automatic certification, and will be subject to a manual review before the container can be approved for publication"
```

dfarrell07:
These EC failures are not related to this change.

aswinsuryan (approved):

tpantelis (approved):

PR #1473, opened 2025-08-08
Test infra EC bug is fixed

dfarrell07:


dfarrell07:
It passed when the infra change was reverted. Trying again now that it's back.

/retest

dfarrell07:
This is fixed in https://github.com/submariner-io/subctl/pull/1474.

PR #1496, opened 2025-09-03
add --skip-intra-cluster-connectivity-tests flag to verify command

yboaron:
Introduce new flag to subctl verify that skips datapath tests involving intra-cluster routing.


cmd/subctl/verify.go
```diff
@@ -132,6 +133,8 @@ func addVerifyFlags(cmd *cobra.Command) {
 	cmd.Flags().UintVar(&verifyFlags.PacketSize, "packet-size", 3000, "set packet size used in TCP connectivity tests")
 	cmd.Flags().BoolVar(&verifyFlags.SkipConnectorSrcIPCheck, "skip-src-ip-check", false,
 		"skip source IP verification for connector pod traffic")
+	cmd.Flags().BoolVar(&verifyFlags.SkipIntraClusterRoutingTests, "skip-intra-cluster-routing-tests", false,
+		"skip tests that verify intra-cluster routing component")
```
tpantelis: I think we should keep the wording consistent with the join flag, ie use the term _connectivity_:

```suggestion
	cmd.Flags().BoolVar(&verifyFlags.SkipIntraClusterRoutingTests, "skip-intra-cluster-connectivity-tests", false,
		"skip tests that verify intra-cluster connectivity functionality")
```

tpantelis (changes requested):

cmd/subctl/verify.go
```diff
@@ -44,6 +44,7 @@ type VerifyOptions struct {
 	VerboseConnectivityVerification bool
 	DisruptiveTests                 bool
 	SkipConnectorSrcIPCheck         bool
+	SkipIntraClusterRoutingTests    bool
```
vthapar: Agree with Tom on using term "Connectivity" over routing. This should also be changed to `SkipIntraClusterConnectivityTests`

cmd/subctl/verify.go
vthapar: @tpantelis `intra-cluster-connecitivity` should be enough, don't need `functionality`

internal/verify/verify.go
```diff
@@ -61,6 +61,7 @@ func runVerify(options subctl.VerifyOptions, fromClusterInfo, toClusterInfo, ext
 	framework.TestContext.SubmarinerNamespace = namespace
 	framework.TestContext.PacketSize = options.PacketSize
 	framework.TestContext.SkipConnectorSrcIPCheck = options.SkipConnectorSrcIPCheck
+	framework.TestContext.SkipIntraClusterRoutingTests = options.SkipIntraClusterRoutingTests
```
vthapar:
```suggestion
	framework.TestContext.SkipIntraClusterConnectivityTests = options.SkipIntraClusterConnectivityTests
```

cmd/subctl/verify.go
vthapar:
```suggestion
	SkipIntraClusterConnectivityTests    bool
```

vthapar (changes requested):

cmd/subctl/verify.go
yboaron: In that case we need to update shipyard code before continuing with this PR, check  https://github.com/submariner-io/shipyard/pull/2119

tpantelis: yeah let's make it consistent. There's also the `SUBMARINER_INTRAROUTINGDISABLED`  env var but I'm fine with leaving that as is since it's internal to the RA.

tpantelis (approved):

vthapar (approved):

PR #1544, opened 2025-10-27
Retrieve the latest release using the HTTP client

skitt:
This avoids a dependency on go-github, which drops a few transitive dependencies that were only used to retrieve the latest release from the GitHub API. This helps our dependency tree...


dfarrell07 (approved):

cmd/subctl/upgrade.go
```diff
@@ -107,12 +112,9 @@ func upgradeSubctl(status reporter.Interface) (string, error) {
...
 	// If the user hasn't specified a version, try to find the latest release on GitHub
 	if upgradeSubctlVersion == "" {
-		client := github.NewClient(nil)
-		latestRelease, _, err := client.Repositories.GetLatestRelease(context.TODO(), "submariner-io", "releases")
-
-		// If we can't determine the latest release, we'll force a download and delegate to get.submariner.io
-		if err == nil {
-			upgradeSubctlVersion = *latestRelease.TagName
+		tag, err := retrieveLatestReleaseTag()
+		if err != nil {
```
tpantelis, line 116: We should at least log the error instead of silencing it. Also shouldn't this be:


```suggestion
		if err == nil {
```

cmd/subctl/upgrade.go
```diff
@@ -165,6 +167,37 @@ func upgradeSubctl(status reporter.Interface) (string, error) {
...
 }
 
+func retrieveLatestReleaseTag() (string, error) {
+	// Retrieve the latest release tag from https://api.github.com/repos/submariner-io/releases/releases/latest
+	// (the duplicate "releases" portion is normal, it points to the releases of the Submariner releases project)
+	httpClient := &http.Client{Timeout: 10 * time.Second}
+
+	r, err := httpClient.Get("https://api.github.com/repos/submariner-io/releases/releases/latest")
+	if err != nil {
+		return "", fmt.Errorf("error accessing GitHub: %w", err)
```
tpantelis, line 179: I think `errors.Wrap` is preferred (here and below).

```suggestion
		return "", errors.Wrap(err, "error accessing GitHub")
```

tpantelis (changes requested):

cmd/subctl/upgrade.go
skitt: Oops, yes, it should be! We didn’t log the error in the past, and where should we log it? This is in `subctl`, so apart from the screen there isn’t anywhere to log it. The error is unlikely to be user-actionable so this would probably only confuse people...

cmd/subctl/upgrade.go
skitt: See https://pkg.go.dev/errors

tpantelis: They're equivalent tho `Wrap` also includes the stack trace. I've found `Wrap` more intuitive. But `github.com/pkg/errors` is no longer maintained and archived so there's that (even tho it's still used). Given that it's fine as is.

cmd/subctl/upgrade.go
tpantelis: We can log it to the `status` as a warning. It may not be actionable but it would be informational, in case something else goes wrong as a side-effect that might provide a clue. I don't think it's best practice to ignore errors.

tpantelis (approved):

PR #1562, opened 2025-11-13
Adjust to golangci-lint 2.6.1 changes

tpantelis:
See commits for details.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -77,9 +77,11 @@ linters:
     wrapcheck:
       extra-ignore-sigs:
         - .Error(
-    wsl:
-      # Separating explicit var declarations by blank lines seems excessive.
-      allow-cuddle-declarations: true
+    wsl_v5:
+      disable:
+        - decl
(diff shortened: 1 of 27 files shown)


.golangci.yml
```diff
@@ -77,9 +77,11 @@ linters:
     wrapcheck:
       extra-ignore-sigs:
         - .Error(
-    wsl:
-      # Separating explicit var declarations by blank lines seems excessive.
-      allow-cuddle-declarations: true
+    wsl_v5:
```
skitt, line 83: The commit message is misleading (`wsl_v6` instead of `wsl_v5`).

skitt (changes requested):
All the authorship info and sign-offs are wrong :wink:.

tpantelis:
> All the authorship info and sign-offs are wrong 😉.

Ugh. I think that happened when I was playing around with trying to get `make release` outside of dapper.

skitt (approved):

PR #1631, opened 2026-02-02
Use Ginkgo context instead of global variables

skitt:
The new `context.TODO()` calls will be addressed in a future change.

## Summary by CodeRabbit

* **Tests**
  * Test suites updated to use explicit per-test contexts across many packages, improving test reliability and consistency.

* **Refactor**
  * Core initialization and helper flows made context-aware and updated to propagate contexts through drivers and helpers, aligning runtime setup and diagnostics with the testing framework while preserving behavior.

### DIFF
--- a/internal/restconfig/restconfig.go
+++ b/internal/restconfig/restconfig.go
@@ -19,6 +19,7 @@ limitations under the License.
 package restconfig
 
 import (
+	"context"
 	goerrors "errors"
 	"fmt"
 	"strings"
@@ -236,7 +237,7 @@ func createClusterInfo(clientConfig clientcmd.ClientConfig, overrides *clientcmd
 		return nil, err
 	}
 
-	return cluster.NewInfo(restConfig.ClusterName, restConfig.Config) //nolint:wrapcheck // No need to wrap
+	return cluster.NewInfo(context.TODO(), restConfig.ClusterName, restConfig.Config) //nolint:wrapcheck // No need to wrap
 }
 
 func (rcp *Producer) runInCluster(function PerContextFn, status reporter.Interface) error {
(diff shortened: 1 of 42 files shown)


internal/restconfig/restconfig.go
```diff
@@ -236,7 +237,7 @@ func createClusterInfo(clientConfig clientcmd.ClientConfig, overrides *clientcmd
 		return nil, err
 	}
 
-	return cluster.NewInfo(restConfig.ClusterName, restConfig.Config) //nolint:wrapcheck // No need to wrap
+	return cluster.NewInfo(context.TODO(), restConfig.ClusterName, restConfig.Config) //nolint:wrapcheck // No need to wrap
```
tpantelis, line 240: Perhaps we should propagate ctx from `createClusterInfo`. But it looks like that would require more refactoring outside the scope of this PR.

pkg/diagnose/kubeproxy_test.go
```diff
@@ -105,5 +107,5 @@ type kubeProxyTestDriver struct {
 }
 
 func (t *kubeProxyTestDriver) run() error {
-	return diagnose.KubeProxyMode(newClusterInfo(), "", []string{}, t.statusTracker)
+	return diagnose.KubeProxyMode(newClusterInfo(context.TODO()), "", []string{}, t.statusTracker)
```
tpantelis, line 110: Same here

tpantelis (approved):

internal/restconfig/restconfig.go
skitt: Yes, see the commit message:

> The new context.TODO() calls will be addressed in a future change.

tpantelis: ok I didn't see that - that message didn't propagate to the PR when you created it.

PR #1761, opened 2026-04-27
Gather nftables rules alongside iptables in subctl gather

yboaron:
Submariner switched from iptables to nftables as the default packet filter driver starting in v0.22.0. However, subctl gather only collects iptables rules, creating a diagnostic gap when analyzing nftables-based deployments offline.

This commit adds collection of nftables rulesets alongside the existing iptables collection:
- nft list ruleset: Complete nftables configuration
- nft -a list ruleset: Same with rule handles (similar to iptables line numbers)

The nftables commands use ignoreError=true to avoid failures on systems that don't have nftables installed, ensuring backward compatibility with iptables-only environments.

## Summary by CodeRabbit

* **New Features**
  * Network diagnostics now collect nftables ruleset output alongside iptables, giving better visibility into active network rules.
* **Bug Fixes**
  * Firewall data collection is more robust: individual nftables query failures are ignored so diagnostics complete reliably.

### DIFF
--- a/internal/gather/cni.go
+++ b/internal/gather/cni.go
@@ -55,6 +55,11 @@ var ipTablesCmds = map[string]string{
 	"iptables-save":   "iptables-save -c",
 }
 
+var nfTablesCmds = map[string]string{
+	"nftables-ruleset": "nft list ruleset",
+	"nftables-save":    "nft -a list ruleset",
+}
+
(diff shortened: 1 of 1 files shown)


tpantelis (approved):

skitt (approved):

yboaron:
@coderabbitai review

PR #1828, opened 2026-07-09
Fix CVEs in release-0.23

vthapar:
See commit messages for details.

## Summary by CodeRabbit

* **Chores**
  * Updated Go module dependency versions (including `golang.org/x/*`) to newer releases.
  * Refreshed vulnerability scan ignore rules, adding specific Go vulnerability exceptions and clarifying related fix-version behavior.


go.mod
```diff
@@ -151,3 +151,7 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.1 // indirect
 )
+
+replace golang.org/x/crypto v0.47.0 => golang.org/x/crypto v0.52.0
+
+replace golang.org/x/net v0.49.0 => golang.org/x/net v0.55.0
```
skitt: This is unnecessary, Go guarantees that the minimum version of a module used to build a binary is the version specified in this `go.mod`. If a security tool complains about a version in `go.sum`, that’s a bug in the security tool, not a problem to be fixed using a `replace` statement.

skitt (changes requested):

go.mod
vthapar: Re-running the skill removed this. It was added coz submariner was usin an older version. Since there was no submariner bump in this set, it likely picked version in submariner from local repo. Either way, would be correct version in all when we release with submariner version bumps.

skitt (changes requested):
The x/sys commit is a downgrade; remove that, and the second x/crypto and x/net commits are no longer necessary. Please also merge the two stdlib commits.

vthapar:
> The x/sys commit is a downgrade; remove that, and the second x/crypto and x/net commits are no longer necessary. Please also merge the two stdlib commits.

Ack. Noticed it after pushing, working on it.

go.mod
skitt: That’s surprising, the skill isn’t supposed to be able to add `replace` directives!

skitt (approved):

PR #1833, opened 2026-07-21
Bump to Kubernetes 1.36 and Go 1.26

skitt:
## Summary by CodeRabbit

* **Maintenance**
  * Updated the Go toolchain and supporting libraries to newer releases.
  * Improved compatibility with current Kubernetes and cloud-provider integrations.
* **Bug Fixes**
  * Improved formatting consistency for status, warning, and failure messages.
  * Preserved accurate security settings for managed workloads.
* **Tests**
  * Updated validation coverage for formatted status messages and diagnostic scenarios.


internal/cli/status_test.go
```diff
@@ -86,7 +86,7 @@ var _ = Describe("Reporter", func() {
 			otherMsg := "Starting something else"
 
 			reporter.Start(startMsg)
-			reporter.Start(otherMsg)
+			reporter.Start("%s", otherMsg)
```
tpantelis, line 89: It's unfortunate we're now forced to use formatting for a non-constant string.  We could add a `Startf`, `Failuref` et al functions.

tpantelis (approved):

